PR #6, opened 2018-03-04
Add mongo/document package which defines methods to Save/Update/Delete documents

### LINKED ISSUE #5: [mongo] There should be a standard way to upsert/update/delete/paranoia-delete documents

Soulou:
Fixes #5


mongo/document/document.go
```diff
@@ -0,0 +1,144 @@
...
+	Document
+	SetUpdatedAt(time.Time)
+}
+
+func Save(ctx context.Context, collectionName string, doc Document) error {
+	log := logger.Get(ctx)
+	c := mongo.Session(log).Clone().DB("").C(collectionName)
+	defer c.Database.Session.Close()
+	log.WithField(collectionName, doc).Debugf("save '%v'", collectionName)
+	_, err := c.UpsertId(doc.GetID(), &doc)
```
EtienneM: Mask the error?

mongo/document/document.go
```diff
@@ -0,0 +1,144 @@
...
+	return err
+}
+
+// Remove the volume from the database.
```
EtienneM: "Remove the volume" -> "Remove the document"

mongo/document/document.go
```diff
@@ -0,0 +1,144 @@
...
+// Handle with care...
+func Destroy(ctx context.Context, collectionName string, doc Document) error {
+	log := logger.Get(ctx)
+	c := mongo.Session(log).Clone().DB("").C(collectionName)
+	defer c.Database.Session.Close()
+	log.WithField(collectionName, doc).Debugf("remove '%v'", collectionName)
+	return c.RemoveId(doc.GetID())
```
EtienneM: Mask the error

mongo/document/document.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func FindOne(ctx context.Context, collectionName string, query bson.M, doc Document) error {
+	log := logger.Get(ctx)
+	c := mongo.Session(log).Clone().DB("").C(collectionName)
+	defer c.Database.Session.Close()
+	return c.Find(query).One(doc)
+}
+
+func WhereParanoia(ctx context.Context, collectionName string, query bson.M, data interface{}) error {
```
EtienneM: Don't you prefer to have a `Where` method which by default exclude documents with a defined `DeletedAt` field?

mongo/document/document.go
Soulou: As it's the only source of error, I actually don't think it's require to mask the error here, with the stacktrace we'll definitely find the source of the error.

mongo/document/document.go
Soulou: As it's the only source of error, I actually don't think it's require to mask the error here, with the stacktrace we'll definitely find the source of the error.

mongo/document/document.go
Soulou: Hmm, no I prefer that Where is doing.. where actually.

PR #7, opened 2018-03-04
Add cronsetup package to interface go-etcd-cron with scalingo's services

Soulou:


cronsetup/cronsetup.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+
+	"github.com/Scalingo/go-etcd-cron"
+	"github.com/Scalingo/go-utils/logger"
+	"github.com/coreos/etcd/clientv3"
+)
+
+type SetupOpts struct {
+	EtcdConfig func() (clientv3.Config, error)
+	Jobs       []etcdcron.Job
```
EtienneM: Shouldn't you import `etcdcron`?

cronsetup/cronsetup.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("fail to get etcd v3 config: %v", err)
+	}
+
+	etcdMutexBuilder, err := etcdcron.NewEtcdMutexBuilder(etcdConfig)
+	if err != nil {
+		return nil, fmt.Errorf("fail to get etcd mutex builder: %v", err)
+	}
+
+	funcCtx := func(ctx context.Context, j etcdcron.Job) context.Context {
```
EtienneM: Is it really what you wanted to do? A function which just do logging?

cronsetup/cronsetup.go
Soulou: the package is imported by `+	"github.com/Scalingo/go-etcd-cron"`

cronsetup/cronsetup.go
Soulou: What else would you like to do?

cronsetup/cronsetup.go
EtienneM: Shouldn't you import it with `etcdcron "github.com/Scalingo/go-etcd-cron"` so that it works?

PR #14, opened 2018-03-05
Add validable interface

### LINKED ISSUE #13: Add a `Validable` interface
With a `Validate(ctx) error` method which will be run before an `Update` or a `Save` call.

johnsudaar:
Fix #13

### DIFF
--- a/mongo/document/base.go
+++ b/mongo/document/base.go
@@ -45,6 +45,6 @@ func (d *Base) destroy(ctx context.Context, collection string) error {
 	return ReallyDestroy(ctx, collection, d)
 }
 
-func (d *Base) Validate(ctx context.Context) *ValidationError {
+func (d *Base) Validate(ctx context.Context) *ValidationErrors {
 	return nil
 }
--- a/mongo/document/base.go
+++ b/mongo/document/base.go
@@ -33,10 +33,6 @@ func (d *Base) ensureCreatedAt() {
 	}
 }
 
-func (d *Base) setCreatedAt(t time.Time) {
-	d.CreatedAt = t
-}
-
 func (d *Base) setUpdatedAt(t time.Time) {
 	d.UpdatedAt = t
 }
@@ -47,4 +43,8 @@ func (d Base) scope(query bson.M) bson.M {
 
 func (d *Base) destroy(ctx context.Context, collection string) error {
 	return ReallyDestroy(ctx, collection, d)
+}
+
+func (d *Base) Validate(ctx context.Context) *ValidationError {
+	return nil
 }
--- a/mongo/document/validation_error.go
+++ b/mongo/document/validation_error.go
@@ -33,3 +33,11 @@ func (v *ValidationError) Set(field, err string) {
 func (v *ValidationError) Get(field string) []string {
 	return v.Errors[field]
 }
+
+func (v *ValidationError) Build() *ValidationError {
+	if len(v.Errors) == 0 {
+		return nil
+	}
+
+	return v
+}
--- a/mongo/document/validation_error.go
+++ b/mongo/document/validation_error.go
@@ -2,17 +2,32 @@ package document
 
 import "bytes"
 
-type ValidationError struct {
+// ValidationErrors is used to store all errors related to the model validation. The typical usecase is:
+//	func (m *MyModel) Validate(ctx context.Context) *ValidationErrors {
+//		validations := document.NewValidationErrors()
+//
+//		if m.Name == "" {
+//			validations.Set("name", "should not be empty")
+//		}
+//
+//		if m.Email == "" {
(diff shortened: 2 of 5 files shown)


mongo/document/validation_error.go
```diff
@@ -0,0 +1,43 @@
+package document
+
+import "bytes"
+
+type ValidationError struct {
+	Errors map[string][]string `json:"errors"`
```
EtienneM: make `Erorrs` private

mongo/document/validation_error.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func NewValidationError() *ValidationError {
```
EtienneM: Don't we want to add an other constructor which takes arugments so that we could write `return NewValidationError("my-field", "errmsg")`?

mongo/document/validation_error.go
```diff
@@ -0,0 +1,43 @@
+package document
+
+import "bytes"
+
+type ValidationError struct {
```
EtienneM: I would prefer calling it `ValidationErrors`

mongo/document/validation_error.go
johnsudaar: This will break the json serialization

mongo/document/validation_error.go
johnsudaar: We could but i don't see what could be the usage

mongo/document/validation_error.go
johnsudaar: :heavy_check_mark:

mongo/document/validation_error.go
EtienneM: When there is only one field to validate?

mongo/document/base.go
```diff
@@ -33,10 +33,6 @@ func (d *Base) ensureCreatedAt() {
 	}
 }
 
-func (d *Base) setCreatedAt(t time.Time) {
```
EtienneM: Woot? Why did you remove that?

mongo/document/validation_error.go
```diff
@@ -0,0 +1,61 @@
...
+			buffer.WriteString(err)
+			buffer.WriteString(", ")
+		}
+	}
+	return buffer.String()
+}
+
+// Set will add an error on a specific field, if the field already contains an error, it will just add it to the current errors list
+func (v *ValidationErrors) Set(field, err string) {
+	v.Errors[field] = append(v.Errors[field], err)
```
EtienneM: Don't we need to initialize  `v.Errors[field]` to an empty slice?

mongo/document/validation_error.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Get will return all errors set for a specific field
+func (v *ValidationErrors) Get(field string) []string {
+	return v.Errors[field]
+}
+
+// Build will send a ValidationErrors struct if there is some errors or nil if no errors has been defined
+func (v *ValidationErrors) Build() *ValidationErrors {
+	if len(v.Errors) == 0 {
```
EtienneM: Shouldn't you be more clever? Like looping through errors to ensure that there is at least an element in an array? I mean, `len(...)` will return 1 for this map:

```go
{
  "my-field": []string{}
}
```
Isn't it?

mongo/document/validation_error.go
EtienneM: No we don't

johnsudaar: Nop, we can use the default 0 value

https://play.golang.org/p/dfhSKeElMDE

mongo/document/base.go
johnsudaar: Not used

mongo/document/validation_error.go
johnsudaar: We could but what's the usecase? How could a user arrive to this position?

EtienneM: As the `Errors` attribute is public, it can happen

johnsudaar: You're righ

mongo/document/validation_error.go
```diff
@@ -41,21 +21,49 @@ func (v *ValidationErrors) Error() string {
...
+//			validations.Set("name", "should not be empty")
+//		}
+//
+//		if m.Email == "" {
+//			validations.Set("email", "should not be empty")
+//		}
+//
+//		return validations.Build()
+//	}
+
```
EtienneM: Remove empty line

mongo/document/base.go
EtienneM: `setUpdatedAt` seems also useless, don't you think?

johnsudaar: Nop it's mendatory to implement the `document` interface.

PR #16, opened 2018-03-05
Add pugblicgen

### LINKED ISSUE #15: New models will include the entire mgo stack when included
To prevent this, we could write a tool that will generate public models from the real ones

johnsudaar:
Fix #15

### DIFF
--- a/publicgen/ast.go
+++ b/publicgen/ast.go
@@ -16,6 +16,7 @@ type field struct {
 	JSONTag string
 	PkgPath string
 	Pointer bool
+	Slice   bool
 }
 
 func newField(field field) *ast.Field {
@@ -36,14 +37,23 @@ func newField(field field) *ast.Field {
 		},
 	}
 
+	fieldAst.Type = ident
+
 	if field.Pointer {
 		fieldAst.Type = &ast.StarExpr{
 			Star: 1,
 			X:    ident,
 		}
-	} else {
-		fieldAst.Type = ident
 	}
+
+	if field.Slice {
+		oldType := fieldAst.Type
+		fieldAst.Type = &ast.ArrayType{
+			Elt:    oldType,
+			Lbrack: 1,
+		}
+	}
+
 	if field.JSONTag != "" {
 		fieldAst.Tag = &ast.BasicLit{
 			Kind:  token.STRING,
@@ -120,15 +130,18 @@ func newAST(packageName string, structs map[string][]field) *ast.File {
 		imports = append(imports, imps...)
 	}
 
-	impAst := newImports(imports)
-
-	return &ast.File{
+	file := &ast.File{
 		Package: 1,
 		Name: &ast.Ident{
 			Name: packageName,
 		},
-		Decls: append([]ast.Decl{impAst}, declsList...),
+		Decls: declsList,
 	}
+
+	if len(imports) > 0 {
+		file.Decls = append([]ast.Decl{newImports(imports)}, file.Decls...)
+	}
+	return file
 }
 
 func writeAst(ast *ast.File, filename string) error {
@@ -146,6 +159,7 @@ func writeAst(ast *ast.File, filename string) error {
 		return err
 	}
 
+	fmt.Println(string(out))
 	err = ioutil.WriteFile(filename, out, 0644)
 	return err
 }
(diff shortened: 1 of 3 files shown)


johnsudaar:
Sample use of this package:
```
package main

import (
	"fmt"
	"os"

	"github.com/Scalingo/go-utils/publicgen"
	"github.com/Scalingo/scalingo-acme-agent/models"
)

func main() {
	err := publicgen.GeneratePublicModels(publicgen.GeneratorParams{
		OutputFile:    "publicapi/publicapi.go",
		OutputPackage: "publicapi",
		Types: []interface{}{
			models.Domain{},
			models.Certificate{},
		},
	})
	if err != nil {
		panic(err)
	}
}
```

johnsudaar:
2018-03-06_1387x1133_scrot

publicgen/reflect.go
```diff
@@ -0,0 +1,80 @@
...
+			continue
+		}
+
+		name := curField.Name
+		if unicode.IsLower([]rune(name)[0]) {
+			continue
+		}
+
+		json := curField.Tag.Get("json")
+		slice := false
```
EtienneM: slice -> `isSlice`?

publicgen/reflect.go
```diff
@@ -0,0 +1,80 @@
...
+		typ := curField.Type
+
+		if curField.Type.Kind() == reflect.Slice {
+			slice = true
+			typ = curField.Type.Elem()
+		}
+
+		pointer := false
```
EtienneM: pointer -> isPointer?

publicgen/reflect.go
```diff
@@ -0,0 +1,80 @@
...
+		name := curField.Name
+		if unicode.IsLower([]rune(name)[0]) {
+			continue
+		}
+
+		json := curField.Tag.Get("json")
+		slice := false
+		typ := curField.Type
+
+		if curField.Type.Kind() == reflect.Slice {
```
EtienneM: curField.Type -> `typ`

publicgen/reflect.go
```diff
@@ -0,0 +1,80 @@
...
+			slice = true
+			typ = curField.Type.Elem()
```
EtienneM: curField.Type -> `typ`

publicgen/reflect.go
```diff
@@ -0,0 +1,80 @@
...
+			typ = typ.Elem()
+			pointer = true
+		}
+
+		pkg := typ.PkgPath()
+		typePrefix := ""
+
+		if pkg != "" {
+			path := strings.Split(pkg, "/")
+			typePrefix = path[len(path)-1] + "."
```
EtienneM: Would be nice to have a comment with an example to better understand this string manipulation

publicgen/reflect.go
```diff
@@ -0,0 +1,80 @@
+package publicgen
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+	"unicode"
+)
+
+var typeStruct = map[string]string{
```
EtienneM: I don't really like the name `typeStruct`. It seems that you use this to map type to the aliased type. Maybe something like `aliasedTypeMap` is better?

publicgen/ast.go
```diff
@@ -0,0 +1,165 @@
...
+
+	return &ast.GenDecl{
+		Rparen: 1,
+		Tok:    token.IMPORT,
+		Specs:  importDecls,
+		Lparen: 1,
+	}
+}
+
+func newAST(packageName string, structs map[string][]field) *ast.File {
```
EtienneM: Add a comment to explain what it does, does not look straight forward

Soulou:
In the last example screenshot, some tags were missing for some `time.Time`, why?

johnsudaar:
They were not any defined in the base model.

Soulou:
So why are there present in the public API, there is something I don't get

johnsudaar:
The original model was like:

```
type Certificate struct{
  ID              bson.ObjectId `json:"id"`
  CreatedAt time.Time        `json:"created_at"`
// ...
  CerticateQueriedAt time.Time
//....
}
```
Note that the json tag is missing on the `CertificateQueriedAt` so i the generator wont add one.

PR #25, opened 2018-03-12
Remove FindSort and FindSortUnscoped, accept variatic arguments in Find, FindUnscoped and FindOne instead

### LINKED ISSUE #24: [document] soft fields management should be simpler, not having specific methods

Soulou:
Fixes #24

### DIFF
--- a/mongo/document/document.go
+++ b/mongo/document/document.go
@@ -117,14 +117,14 @@ func find(ctx context.Context, collectionName string, query bson.M, doc interfac
 	return c.Find(query).Sort(fields...).One(doc)
 }
 
-func Where(ctx context.Context, collectionName string, query bson.M, data interface{}) error {
+func Where(ctx context.Context, collectionName string, query bson.M, data interface{}, sortFields ...SortField) error {
 	if query == nil {
 		query = bson.M{}
 	}
 	if _, ok := query["deleted_at"]; !ok {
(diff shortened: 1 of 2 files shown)


Soulou:
Would you rather something like:

```
document.Find(ctx, name, id, &doc, document.SortField{"name"})
```

?

mongo/document/document.go
```diff
@@ -142,25 +136,29 @@ func WhereSortUnscoped(ctx context.Context, collectionName string, query bson.M,
...
+	}
+	err := c.Find(query).Sort(fields...).All(data)
 	if err != nil {
 		return fmt.Errorf("fail to query mongo %v: %v", query, err)
 	}
 	return nil
 }
 
-func WhereSort(ctx context.Context, collectionName string, query bson.M, data interface{}, sortFields ...string) error {
+func WhereSort(ctx context.Context, collectionName string, query bson.M, data interface{}, sortFields ...SortField) error {
```
johnsudaar: This is not needed anymore, we could simply use the Where method

Soulou: Yes true, thanks

Soulou: updated

johnsudaar:
The `Where` method is missing the sort fields.

Soulou:
Done

PR #28, opened 2018-06-04
Query builder use a pointer to a query

EtienneM:
So that we don't need to call:

```
query = query.On(...)
```

We just need:

```
query.On(...)
```

### DIFF
--- a/influx/querybuilder.go
+++ b/influx/querybuilder.go
@@ -55,8 +55,8 @@ type conditionOperator struct {
 
 // NewQuery makes a new Query object. You MUST use this method to instantiate a new Query
 // structure.
-func NewQuery() Query {
-	return Query{
+func NewQuery() *Query {
+	return &Query{
 		conditions: nil,
 		limit:      -1,
 	}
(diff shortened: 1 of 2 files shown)


Soulou:
I'm not sure I agree with this, I like the stateless model, you have a query, then you modify it, and tadam you get a new query.

EtienneM:
I don't really see the pros of a stateless query here.

Moreover, we currently have a mix of stateless and stateful. The conditions are stateful as we store a pointer to condition in the Query struct:

```
query := ....
fmt.Println("BINIOU:", query.Build())
query.And("value", influx.Equal, "mean")
fmt.Println("AND:", query.Build())
```

```
BINIOU: SELECT FROM  WHERE time < now() - 1m
AND: SELECT FROM  WHERE time < now() - 1m AND value = mean
```

johnsudaar:
You're right, but i tend to agree with @Soulou on this one and prefer the stateless model. I think that the condition pointer issue is a bug.

EtienneM:
> You're right, but i tend to agree with @Soulou on this one and prefer the stateless model

OK but why? I am just wondering why the stateless model is preferable here?

Soulou:
@EtienneM in a builder like this, I would say that it's less error-prone to have only stateless operations. If there are any stateful ones, it should be fixed. It's really more difficult to read and track when there is a state. Today I think the less states the better =D

EtienneM:
OK then I close this PR in favor of a new one to remove the statefulness of the query builder

PR #34, opened 2018-06-12
Load balance nsq producers

### LINKED ISSUE #23: Add a ConfigFromEnv method for NSQ
Just like the etcd.ConfigFromEnv

### LINKED ISSUE #33: Load balance nsq message accross multiple hosts

johnsudaar:
Fix #33
Fix #23

### DIFF
--- a/nsqlbproducer/lbproducer.go
+++ b/nsqlbproducer/lbproducer.go
@@ -1,4 +1,4 @@
-package nsqproducer
+package nsqlbproducer
 
 import (
 	"context"
--- a/nsqlbproducer/lbproducer.go
+++ b/nsqlbproducer/lbproducer.go
@@ -4,13 +4,16 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"time"
 
+	"github.com/Scalingo/go-utils/nsqproducer"
 	nsq "github.com/nsqio/go-nsq"
 	errgo "gopkg.in/errgo.v1"
 )
 
 type NsqLBProducer struct {
-	producers []Producer
+	producers []nsqproducer.Producer
+	randSrc   randSource
 }
 
 type Host struct {
@@ -24,16 +27,20 @@ type LBProducerOpts struct {
 	SkipLogSet map[string]bool
 }
 
-func NewLB(opts LBProducerOpts) (*NsqLBProducer, error) {
+type randSource interface {
+	Int() int
+}
+
+func New(opts LBProducerOpts) (*NsqLBProducer, error) {
 	if len(opts.Hosts) == 0 {
 		return nil, fmt.Errorf("A producer must have at least one host")
 	}
 	producer := &NsqLBProducer{
-		producers: make([]Producer, len(opts.Hosts)),
+		producers: make([]nsqproducer.Producer, len(opts.Hosts)),
 	}
 
 	for i, h := range opts.Hosts {
-		p, err := New(ProducerOpts{
+		p, err := nsqproducer.New(nsqproducer.ProducerOpts{
 			Host:       h.Host,
 			Port:       h.Port,
 			NsqConfig:  opts.NsqConfig,
@@ -47,11 +54,13 @@ func NewLB(opts LBProducerOpts) (*NsqLBProducer, error) {
 		producer.producers[i] = p
 	}
 
+	producer.randSrc = rand.New(rand.NewSource(time.Now().Unix()))
+
 	return producer, nil
 }
 
-func (p *NsqLBProducer) Publish(ctx context.Context, topic string, message NsqMessageSerialize) error {
-	firstProducer := rand.Int() % len(p.producers)
+func (p *NsqLBProducer) Publish(ctx context.Context, topic string, message nsqproducer.NsqMessageSerialize) error {
+	firstProducer := p.randSrc.Int() % len(p.producers)
 
 	var err error
 	for i := 0; i < len(p.producers); i++ {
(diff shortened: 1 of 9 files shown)


nsqlbproducer/lbproducer.go
```diff
@@ -0,0 +1,94 @@
...
+
+type Host struct {
+	Host string
+	Port string
+}
+
+type LBProducerOpts struct {
+	Hosts      []Host
+	NsqConfig  *nsq.Config
+	SkipLogSet map[string]bool
```
EtienneM: Can you add a comment to explain what this is.

nsqlbproducer/lbproducer.go
```diff
@@ -0,0 +1,94 @@
...
+	"context"
+	"fmt"
+	"math/rand"
+	"time"
+
+	"github.com/Scalingo/go-utils/nsqproducer"
+	nsq "github.com/nsqio/go-nsq"
+	errgo "gopkg.in/errgo.v1"
+)
+
```
EtienneM: Can you add a comment to explain the purpose of this package?

nsqlbproducer/lbproducer.go
```diff
@@ -0,0 +1,94 @@
...
+	for i, h := range opts.Hosts {
+		p, err := nsqproducer.New(nsqproducer.ProducerOpts{
+			Host:       h.Host,
+			Port:       h.Port,
+			NsqConfig:  opts.NsqConfig,
+			SkipLogSet: opts.SkipLogSet,
+		})
+
+		if err != nil {
+			return nil, errgo.Mask(err)
```
EtienneM: I think that we are now using the package`errors`

Soulou:
LGTM

nsqlbproducer/lbproducer.go
```diff
@@ -0,0 +1,94 @@
...
+
+	return producer, nil
+}
+
+func (p *NsqLBProducer) Publish(ctx context.Context, topic string, message nsqproducer.NsqMessageSerialize) error {
+	firstProducer := p.randSrc.Int() % len(p.producers)
+
+	var err error
+	for i := 0; i < len(p.producers); i++ {
+		err = p.producers[(i+firstProducer)%len(p.producers)].Publish(ctx, topic, message)
```
EtienneM: Do we really want to ignore the errors? I mean if there are multiple producers and one fail, we will not be notified about this error

nsqlbproducer/lbproducer.go
```diff
@@ -0,0 +1,94 @@
...
+	"math/rand"
+	"time"
+
+	"github.com/Scalingo/go-utils/nsqproducer"
+	nsq "github.com/nsqio/go-nsq"
+	errgo "gopkg.in/errgo.v1"
+)
+
+type NsqLBProducer struct {
+	producers []nsqproducer.Producer
```
EtienneM: Can you add something to ensure `NsqLBProducer` implements all `nsqproducer.Producer` methods?

nsqproducer/env.go
```diff
@@ -0,0 +1,56 @@
...
+
+	return New(ProducerOpts{
+		Host: E["NSQD_TLS"],
+		Port: E["NSQD_PORT"],
+
+		NsqConfig: nsqConfig,
+	})
+}
+
+func NsqConfigFromEnv(E map[string]string) (*nsq.Config, error) {
```
EtienneM: This function can be private

nsqproducer/producer.go
```diff
@@ -16,6 +16,7 @@ import (
 type Producer interface {
 	Publish(ctx context.Context, topic string, message NsqMessageSerialize) error
 	DeferredPublish(ctx context.Context, topic string, delay int64, message NsqMessageSerialize) error
+	Stop()
```
EtienneM: Shouldn't you implement this method in the `NsqProducer` (non-LB)?

nsqlbproducer/lbproducer.go
johnsudaar: Yeah that's the point, if a a producer get the message it's good enough.
Monitoring the cluster health is another issue which is not in the scope of this PR IMHO.

nsqproducer/env.go
johnsudaar: Nop, it's used by nsqlbproducer.

nsqproducer/producer.go
johnsudaar: It's already there.

nsqlbproducer/lbproducer.go
EtienneM: Do you now what are the reason for a failed publish? How could we monitor such problem elsewhere than with this error?

johnsudaar: A node is down (or is being upgraded) we could monitor it with Philae (having one probe per member of the cluster).

EtienneM: Why not. I am not a big fan of ignoring such errors but OK as it didn't hurt our beloved CTO

Soulou: What about logging all the errors?

johnsudaar: The most common case of this error will be node reboot, nsq reboot, and nsq update. Which are not critical nor important. The only important thing is that the message is sent. So if all node do not respond thats bad, but if one node does not we don't care.

johnsudaar: I do not like it here since this is a library, logging might be hard since we do not know how the user might format his logs.

johnsudaar: You meant logrus.Logger ?

Soulou: logrus.FieldLogger actually

PR #47, opened 2018-07-02
prevent panic if ValidationErrors is nil

### LINKED ISSUE #46: ValidationErrors.Error should not panic with a nil value

EtienneM:
Fix #46

### DIFF
--- a/mongo/document/validation_error.go
+++ b/mongo/document/validation_error.go
@@ -8,6 +8,9 @@ type ValidationErrors struct {
 }
 
 func (v *ValidationErrors) Error() string {
+	if v == nil {
+		return ""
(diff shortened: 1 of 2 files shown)


mongo/document/validation_error.go
```diff
@@ -8,6 +8,9 @@ type ValidationErrors struct {
 }
 
 func (v *ValidationErrors) Error() string {
+	if v == nil {
+		return ""
```
Soulou: `return "empty validation error"` ?

LIke that it's easier to track if we get this error?

EtienneM:
A library can panic, no problem with that. Just use it properly

PR #49, opened 2018-08-13
Add EnsureParanoidIndices method

### LINKED ISSUE #12: Missing index on the `deleted_at` field
Wow how could you forget such a basic feature...

:heart:  @Soulou

EtienneM:
Is it what you meant @johnsudaar ?

Fix #12

### DIFF
--- a/mongo/document/base.go
+++ b/mongo/document/base.go
@@ -13,7 +13,7 @@ type Base struct {
 	UpdatedAt time.Time     `bson:"updated_at" json:"updated_at"`
 }
 
-func (d *Base) IsPersisted() bool {
+func (d Base) IsPersisted() bool {
 	return !d.CreatedAt.IsZero()
 }
 
@@ -45,6 +45,6 @@ func (d *Base) destroy(ctx context.Context, collection string) error {
 	return ReallyDestroy(ctx, collection, d)
 }
 
-func (d *Base) Validate(ctx context.Context) *ValidationErrors {
+func (d Base) Validate(ctx context.Context) *ValidationErrors {
 	return nil
 }
--- a/mongo/document/base.go
+++ b/mongo/document/base.go
@@ -13,11 +13,11 @@ type Base struct {
 	UpdatedAt time.Time     `bson:"updated_at" json:"updated_at"`
 }
 
-func (d Base) IsPersisted() bool {
+func (d *Base) IsPersisted() bool {
 	return !d.CreatedAt.IsZero()
 }
 
-func (d Base) getID() bson.ObjectId {
+func (d *Base) getID() bson.ObjectId {
 	return d.ID
 }
 
--- a/mongo/document/base.go
+++ b/mongo/document/base.go
@@ -45,6 +45,6 @@ func (d *Base) destroy(ctx context.Context, collection string) error {
 	return ReallyDestroy(ctx, collection, d)
 }
 
-func (d Base) Validate(ctx context.Context) *ValidationErrors {
+func (d *Base) Validate(ctx context.Context) *ValidationErrors {
 	return nil
 }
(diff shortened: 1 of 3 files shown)


mongo/document/base.go
```diff
@@ -13,11 +13,11 @@ type Base struct {
 	UpdatedAt time.Time     `bson:"updated_at" json:"updated_at"`
 }
 
-func (d Base) IsPersisted() bool {
```
Soulou: Nope, no need of pointer, nothing is modified

mongo/document/base.go
```diff
@@ -13,11 +13,11 @@ type Base struct {
...
+func (d *Base) IsPersisted() bool {
 	return !d.CreatedAt.IsZero()
 }
 
-func (d Base) getID() bson.ObjectId {
```
Soulou: No need of pointer, nothing is modified

mongo/document/paranoid.go
```diff
@@ -12,15 +12,15 @@ type Paranoid struct {
 	DeletedAt *time.Time `bson:"deleted_at,omitempty" json:"deleted_at,omitempty"`
 }
 
-func (p Paranoid) IsDeleted() bool {
```
Soulou: No pointer

mongo/document/paranoid.go
```diff
@@ -12,15 +12,15 @@ type Paranoid struct {
...
+func (p *Paranoid) IsDeleted() bool {
 	return p.DeletedAt != nil
 }
 
 func (d *Paranoid) setDeletedAt(t time.Time) {
 	d.DeletedAt = &t
 }
 
-func (d Paranoid) scope(query bson.M) bson.M {
```
Soulou: No pointer

mongo/document/document.go
```diff
@@ -22,14 +24,23 @@ type document interface {
 	Validable
 }
 
+var _ document = &Base{}
```
Soulou: No pointer

mongo/document/document.go
```diff
@@ -221,3 +234,26 @@ func Update(ctx context.Context, collectionName string, update bson.M, doc docum
 	log.WithField("query", update).Debugf("update %v", collectionName)
 	return c.UpdateId(doc.getID(), update)
 }
+
+func EnsureParanoidIndices(ctx context.Context, collectionNames ...string) {
+	go func() {
```
Soulou: Why in a goroutine?

mongo/document/base.go
EtienneM: Then document `Base` does not satisfy the `document` interface as `getID` is implemented for `Base` but the `ensure*` methods are for pointer of Base.

EtienneM:
I think it's mandatory to have everything as a point. For instance, the `document` interface requires, among other, the following two methods:

```go
setUpdatedAt(time.Time)
getID() bson.ObjectId
```

The `Base` structure implements this interface. `setUpdatedAt` requires to modify the `Base` instance. `getID` does not. But if I the receiver of `setUpdatedAt` is a pointer and `getID` is not, `Base` does not implement the `document` interface (https://play.golang.org/p/CqjoJPV3b8d).

mongo/document/document.go
EtienneM: The `Base` implementation of the `document` interface is on pointer. Cf. https://github.com/Scalingo/go-utils/pull/49/files#diff-9732770faadda6716a617078c80ccbe9L24

mongo/document/base.go
johnsudaar: Go is smarter than that :P If the struct is passed as a pointer it can access non-pointer methods: https://play.golang.com/p/NicC6AlMbd7

(Not the over way around though: https://play.golang.com/p/o-eDYPih2V4)

EtienneM: Yep but interface does not work that way. If you have the following struct and interface:

```go
type T struct {}

type I interface {
  A()
  B()
}
```

And you want the struct `T` to implement the interface `I`, you can NOT do the following:

```go
func (t T) A() {}
func (t *T) B() {}
```

With this example, the struct `T` does not satisfy the interface `I`

PR #58, opened 2018-10-10
Add httpclient lib forwarding the request_id from context

Soulou:


httpclient/httpclient.go
```diff
@@ -0,0 +1,80 @@
...
+		parent := &http.Transport{
+			TLSClientConfig: config,
+		}
+		c.Transport = reqidTransport{parent: parent}
+	}
+}
+
+func WithAuthentication(username, password string) ClientOpt {
+	return func(c *http.Client) {
+		c.Transport = authTransport{
```
EtienneM: Don't you override this Transport if you want the client to use both `WithTLSConfig` and `WithAuthentication`?

Soulou: Updated

PR #75, opened 2019-03-05
[mongodb] Add a default connection timeout of 10 seconds

Soulou:
That's what is done by default in:

    mgo.Dial() -> internals are adding a timeotu of 10 seconds

### DIFF
--- a/mongo/mongo.go
+++ b/mongo/mongo.go
@@ -57,6 +57,7 @@ func BuildSession(ctx context.Context, rawURL string) (*mgo.Session, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


mongo/mongo.go
```diff
@@ -57,6 +57,7 @@ func BuildSession(ctx context.Context, rawURL string) (*mgo.Session, error) {
 	if err != nil {
 		return nil, err
 	}
+	info.Timeout = 10 * time.Second
```
EtienneM: Do we want to add a way to specify an argument in the connection string to customize this? Such as `mongodb://...?timeout=5s`

Soulou: Done

PR #76, opened 2019-03-05
The ? mark should not be removed when parsing ssl/timeout

Soulou:
`mongodb://localhost:27017/db?ssl=true&option=value`

should not become

    mongodb://localhost:27017/db&option=value

but

    mongodb://localhost:27017/db?&option=value

which is valid


EtienneM:
Don't forget to backport it to the app-scheduler

Soulou:
https://github.com/Scalingo/app-scheduler/pull/457

PR #80, opened 2019-04-02
Add support for subqueries

### LINKED ISSUE #79: [InfluxDB Query Builder] Implement subqueries
https://docs.influxdata.com/influxdb/v1.7/query_language/data_exploration/#subqueries

EtienneM:
Fix #79

### DIFF
--- a/influx/execute.go
+++ b/influx/execute.go
@@ -8,7 +8,7 @@ import (
 )
 
 // Do actually executes the query to the specified InfluxDB instance hosted at the url argument.
-func (q Query) Do(url string) (*influx.Response, error) {
+func (q *Query) Do(url string) (*influx.Response, error) {
 	query := q.Build()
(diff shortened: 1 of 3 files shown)


influx/execute.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 // Do actually executes the query to the specified InfluxDB instance hosted at the url argument.
-func (q Query) Do(url string) (*influx.Response, error) {
+func (q *Query) Do(url string) (*influx.Response, error) {
```
johnsudaar: The fact that the query was not a pointer was by design. It let us reuse the query builder.

```
a := NewQuery().On("").Limit().GroupBy() ...

web := a.Where("container_type=web")
clock := a.Where("container_type=clock")
```

EtienneM: OK I see the use case and the pros

PR #83, opened 2019-04-12
Better handling of ValidationErrors

EtienneM:
Related to Scalingo/go-handlers#15

### DIFF
--- a/errors/cause.go
+++ b/errors/cause.go
@@ -1,0 +1,21 @@
+package errors
+
+import (
+	"reflect"
+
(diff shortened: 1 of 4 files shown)


EtienneM:
Example of use here (https://github.com/Scalingo/go-handlers/pull/16/files#diff-d5c648cce93d1693cc30823c479e25dbR57) and there (https://github.com/Scalingo/go-handlers/pull/16/files#diff-d5c648cce93d1693cc30823c479e25dbR69)

PR #89, opened 2019-07-19
Add simple object storage package for s3 and swift

Soulou:


storage/s3.go
```diff
@@ -0,0 +1,135 @@
...
+	s3uploader        *s3manager.Uploader
+	retryWaitDuration time.Duration
+	retryAttempts     int
+}
+
+func NewS3(cfg S3Config, bucket string) *S3 {
+	s3config := s3Config(cfg)
+	return &S3{
+		cfg: cfg, s3client: s3.New(s3config), s3uploader: s3manager.NewUploader(s3config),
+		retryWaitDuration: time.Second, retryAttempts: 3,
```
johnsudaar: Shouldn't those two fields be configurable ?

Soulou: Actually after thinking I'm wondering if it's not the package use who should be responsible of retrying a Size() command (waiting for eventual consistency right after an upload) What do you think?

Soulou:
> It would be nice if the common interface and it's mock would be in the package too.

Ok will do

storage/s3.go
johnsudaar: mmmh if the service is eventually consistent i find it pretty nice that the package offer this option. Otherwise it should be configurable.

Soulou: Hmmm ok :-)

Soulou:
mock added

PR #94, opened 2019-11-12
Deps: update to new location of etcd client dependency go.etcd.io/etcd with its dependencies

Soulou:


Soulou:
I'm in the process of updating sand that's why I've to update this..

PR #103, opened 2020-02-21
[logger] Add default field: region if environment REGION_NAME is defined

Soulou:


Soulou:
There is no choice here if we want this kind of features to be added transparently.. No way to add a field on a `*logurs.Logger`

PR #110, opened 2020-03-27
[retry] Allow stopping retry with a custom option: WithMaxDuration

Soulou:


Soulou:
It allows me to write this, 

```
	retrier := retry.New(
		retry.WithWaitDuration(client.containerStateCheckInterval),
		retry.WithTimeout(client.containerDeathTimeout),
	)
	err := retrier.Do(ctx, func(ctx context.Context) error {
```
instead of this:

```
	retrier := retry.New(
		retry.WithWaitDuration(client.containerStateCheckInterval),
		retry.WithMaxAttempts(int(client.containerDeathTimeout/client.containerStateCheckInterval)+10),
	)
	ctx, cancel := context.WithTimeout(ctx, client.containerDeathTimeout)
	defer cancel()

	err := retrier.Do(ctx, func(ctx context.Context) error {
```

Which is according to me much more elegant

retry/retry.go
```diff
@@ -54,11 +81,10 @@ func (r Retryer) Do(ctx context.Context, method Retryable) error {
 		timer := time.NewTimer(r.waitDuration)
 		select {
 		case <-timer.C:
+		case <-timeoutCtx.Done():
+			return timeoutCtx.Err()
```
johnsudaar: You should specialize the context errors here.

Soulou: Done

retry/retry.go
```diff
@@ -2,9 +2,27 @@ package retry
...
 )
 
+type RetryErrScope string
+
+const (
+	MaxDurationScope RetryErrScope = "max-duration"
+	ContextScope     RetryErrScope = "context"
+)
+
+type RetryErr struct {
```
EtienneM: Following this golang doc (https://github.com/golang/go/wiki/Errors#naming), you should name this:
```suggestion
type RetryError struct {
```

Soulou: Done

PR #120, opened 2020-04-26
Externalize 'fs' package from private project

Soulou:
Needed for https://github.com/Scalingo/go-utils/pull/121

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -1,0 +1,55 @@
+package fs
+
+import (
+	"os"
+
+	"github.com/spf13/afero"
+)
+
+// Fs interface contains all methods implemented in MemFs and OsFs
+type Fs interface {
+	afero.Fs
+	Chown(string, int, int) error
+	Link(string, string) error
+	Symlink(string, string) error
+}
+
+type MemFs struct {
+	afero.Fs
+}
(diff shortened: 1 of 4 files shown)


fs/fs.go
```diff
@@ -0,0 +1,55 @@
...
+func (fs MemFs) Link(string, string) error {
+	return nil
+}
+
+func (fs MemFs) Symlink(string, string) error {
+	return nil
+}
+
+type OsFs struct {
+	afero.Fs
```
EtienneM, line 38: I don't understand where the `afero.Fs` attribute is used. Why isn't it used in the various implemented methods?

fs/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/Scalingo/go-utils/fs
+
+go 1.13
```
EtienneM, line 3:
```suggestion
go 1.14
```

fs/fs.go
Soulou: It's because afero.Fs is already implementing a  lot of methods, we're just adding a few here.

```
osfs := fs.NewOsFs()
osfs.Mkdir() <- using afero.Fs
```

PR #121, opened 2020-04-26
Externalize 'tarball' package from private project

Soulou:


tarball/go.mod
```diff
@@ -0,0 +1,16 @@
...
+go 1.13
+
+require (
+	github.com/Scalingo/go-utils v6.4.0+incompatible
+	github.com/Scalingo/go-utils/fs v0.0.0-20200427101022-73f3c803d841
+	github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
+	github.com/pkg/errors v0.9.1
+	github.com/sirupsen/logrus v1.5.0 // indirect
+	github.com/spf13/afero v1.2.2
+	github.com/stretchr/testify v1.2.2
```
EtienneM, line 12: Update to the latest 1.5.1? We will forget to update it in the future so at least use the latest version at the time of release :P

tarball/archive.go
```diff
@@ -0,0 +1,175 @@
...
+	"strconv"
+	"strings"
+
+	fspkg "github.com/Scalingo/go-utils/fs"
+	"github.com/Scalingo/go-utils/logger"
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
+)
+
+var EmptyArchiveError = errors.New("empty archive")
```
EtienneM, line 20:
```suggestion
var ErrEmptyArchive = errors.New("empty archive")
```

Error variables start with `Err` (https://github.com/golang/go/wiki/Errors#naming) :grin:

Soulou:
copy pasted with some additional specs ;-)

tarball/archive.go
Soulou: deleted: was not used

tarball/go.mod
Soulou: Done

PR #128, opened 2020-08-13
[Document] Do not add sensitive information in the log

EtienneM:


mongo/document/document.go
```diff
@@ -65,7 +65,7 @@ func Create(ctx context.Context, collectionName string, doc document) error {
 
 	c := mongo.Session(log).Clone().DB("").C(collectionName)
 	defer c.Database.Session.Close()
-	log.WithField(collectionName, doc).Debugf("save '%v'", collectionName)
+	log.WithField("collection", collectionName).Debugf("save '%v'", collectionName)
```
johnsudaar, line 68: Could you add the Document ID ? This is not sensitive information and can be very useful for debugging.

graceful/service_test.go
```diff
@@ -161,7 +161,9 @@ func ensurePidFileProcessKilled(t *testing.T) {
 		process, err := os.FindProcess(pid)
 		require.NoError(t, err)
 		err = process.Kill()
-		require.NoError(t, err)
+		if err != nil && !strings.Contains(err.Error(), "already finished") {
```
johnsudaar, line 164: Have you checked with @Soulou if this is normal ? This seems weird that the process has already exited.

Soulou: Hmm yes I don't really get this =\

EtienneM:
> Weird change for the graceful spec, except this I'm ok

Yes but the specs were failing (https://travis-ci.com/github/Scalingo/go-utils/builds/179556969) but I couldn't figure out why...

PR #133, opened 2020-10-07
Add package 'concurrency' with struct ParallelWorker, include doc and example as test

Soulou:
FYI 'go test' execute the Example and compare the output as the comment // Output: [...]

### DIFF
--- a/concurrency/parallel_worker.go
+++ b/concurrency/parallel_worker.go
@@ -1,0 +1,47 @@
+package concurrency
+
+import "sync"
+
+// ParallelWorker is a struct providing an helper to run tasks in parallel
+// while capping the parallelism with a known number of workers.
+type ParallelWorker struct {
+	sem         chan struct{}
+	wg          *sync.WaitGroup
+	endFunction func()
+}
+
+// NewParallelWorker constructs a new parallel worker running at maximum
+// <workers> jobs at a time. <endFunc> is a callback called when all the jobs
+// are over and that "Stop" is called.
+func NewParallelWorker(workers int, endFunc func()) ParallelWorker {
+	w := ParallelWorker{
+		sem:         make(chan struct{}, workers),
+		wg:          &sync.WaitGroup{},
+		endFunction: endFunc,
+	}
+
+	return w
+}
+
+// Stop should be called to release all resources. It waits for all the runnings
+// tasks to be over Calling Perform after this would result in a panic
+func (w ParallelWorker) Stop() {
+	w.wg.Wait()
+	close(w.sem)
+	w.endFunction()
+}
+
+// Perform adds a job to handle. It can be called any number of time The
+// concurrencry will never get over the number of workers defined at the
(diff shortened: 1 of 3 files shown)


concurrency/parallel_worker.go
```diff
@@ -0,0 +1,47 @@
...
+		wg:          &sync.WaitGroup{},
+		endFunction: endFunc,
+	}
+
+	return w
+}
+
+// Stop should be called to release all resources. It waits for all the runnings
+// tasks to be over Calling Perform after this would result in a panic
+func (w ParallelWorker) Stop() {
```
EtienneM, line 28: I don't find the function name good. I think it does not reflect the fact that it's actually a wait. Maybe `WaitAndDone` or `WaitAndFinish`?

concurrency/parallel_worker.go
```diff
@@ -0,0 +1,47 @@
...
+	w.wg.Wait()
+	close(w.sem)
+	w.endFunction()
+}
+
+// Perform adds a job to handle. It can be called any number of time The
```
EtienneM, line 34:
```suggestion
// Perform adds a job to handle. It can be called any number of times. The
```

concurrency/parallel_worker.go
```diff
@@ -0,0 +1,47 @@
...
+// concurrencry will never get over the number of workers defined at the
```
EtienneM, line 35:
```suggestion
// concurrency will never get over the number of workers defined at the
```

EtienneM (commented):
Out of curiosity, what is the use case you are working on? :)

concurrency/parallel_worker.go
Soulou: ACK

Soulou:
> Out of curiosity, what is the use case you are working on? :)

It's about the `scalingo-overview` CLI I've shown in random ;-)

concurrency/parallel_worker.go
Soulou: @EtienneM would `Complete()` would be better? =p

EtienneM: I'm not sure. The `Stop` method is really just a wait operation. And from my perspective, `Complete` is not clear about the face that the method will block until the computation is over.

Soulou: I've used `CompleteProcessing`

In English it sounds semantically quite good to me, what do you think?

EtienneM: Yep OK, thanks :)

EtienneM (approved):
Fix the specs then LGTM

Soulou:
@EtienneM specs are failing because of another package I would be really grateful if it would be merged =P

Soulou:
(one build succeeds though!)

EtienneM:
OK, I opened an issue about these random specs (https://github.com/Scalingo/go-utils/issues/134).

PR #137, opened 2020-11-23
[gomock_generator] go modules compatibility, pretty JSON and bugfix

Soulou:


Soulou:
Used in https://github.com/Scalingo/appsdeck-database/pull/1886/commits/165447f419ee1d693dd30ac761915ee035bc81f2

gomock_generator/gomockgenerator/generate.go
```diff
@@ -195,10 +205,15 @@ func generateMock(ctx context.Context, basePackage string, mock MockConfiguratio
 		"current":  hash,
 	}).Info("Signature is not matching, regenerating")
 
+	gomod := "--build_flags=--mod=mod"
```
johnsudaar, line 208: Couldn't this be dangerous ? Couldn't gomock generate a file that does not correspond to the hash calculated ? (Since the hash comes from the filesystem and the hash from the filesystem). Shouldn't we use `--mod=vendor` instead ?

Soulou: Sometimes you want to generate mocks for libs, in this case, you don't want to use `--mod=vendor`

When using `--mod=mod` either the right version is on the FS `$GOPATH/pkg/mod/...` or in the vendor dir, in both cases, it's the same file found by go modules.

johnsudaar: Well if it's for a lib, the lib is supposed to be in the vendor directory no ?

Soulou: But when you're working on a lib, and on specs of a lib, you might need to create mocks, and there is no vendor directory

johnsudaar: Ahhh ok LGTM

johnsudaar (approved):

PR #140, opened 2021-01-11
[Mongo] Add scalingo pagination support to the package

curzolapierre:
A pagination method to follow [scalingo pagination](https://developers.scalingo.com/index#pagination)

Related to [GRR-81]

[GRR-81]: https://scalingo.atlassian.net/browse/GRR-81

### DIFF
--- a/mongo/pagination/README.md
+++ b/mongo/pagination/README.md
@@ -24,14 +24,18 @@ Then call the `Paginate` as follow:
 resultObject := []*ResultObject{}
 dbQuery := bson.M{"searched_field": "field_1"}
 
+paginateOpts := PaginateOpts{
+    PageNumber:  5,         // Number of the requested page (can be empty, default 1)
+    AmountItems: 10,        // Amount of items by page (can be empty)
+    Query:       dbQuery,   // Query which will be executed on the database (can be nil)
+    SortOrder:   "-_id",    // The field for the sort order (by default "_id")
+}
+
 meta, err := pageService.Paginate(
     ctx,                    // A context (required)
-    "5",                    // Number of the requested page (can be empty, default 1)
-    "10",                   // Amount of items by page (can be empty)
-    dbQuery,                // Query which will be executed on the database (can be nil)
     "RequestedCollection",  // Name of the collection (required)
     &resultObject,          // The object that will contain the data (must be an array)
-    "-_id")                 // The field for the sort order (required)
+    paginateOpts)
 ```
 
 The returned meta object contains pagination metadata that could be used in the
--- a/mongo/pagination/README.md
+++ b/mongo/pagination/README.md
@@ -1,0 +1,38 @@
+# Mongo Pagination
+
+Pagination is a package providing a function (`Paginate`) that queries the mongo
+database and return results by pages. It is useful to avoid the processing of
+too much data.
+
+The pagination follow the rules explained in the
+[Scalingo pagination documentation](https://developers.scalingo.com/index#pagination)
+
+## How to use it
+
+First, set the default value of number of items per page and the maximum of item
+by page as the following snippet:
+
+```Go
+pageService := pagination.NewPaginationService(pagination.ServiceOpts{
+	PerPageDefault: 5,
+	MaxPerPage:     15,
+})
+```
+
+Then call the `Paginate` as follow:
+```Go
+resultObject := []*ResultObject{}
+dbQuery := bson.M{"searched_field": "field_1"}
+
+meta, err := pageService.Paginate(
+    ctx,                    // A context (required)
+    "5",                    // Number of the requested page (can be empty, default 1)
+    "10",                   // Amount of items by page (can be empty)
+    dbQuery,                // Query which will be executed on the database (can be nil)
+    "RequestedCollection",  // Name of the collection (required)
+    &resultObject,          // The object that will contain the data (must be an array)
+    "-_id")                 // The field for the sort order (required)
+```
+
+The returned meta object contains pagination metadata that could be used in the
+request answer.
--- a/mongo/pagination/pagination.go
+++ b/mongo/pagination/pagination.go
@@ -3,7 +3,6 @@ package pagination
 import (
 	"context"
 	"fmt"
-	"net/url"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


mongo/pagination/pagination.go
```diff
@@ -0,0 +1,172 @@
...
+	CurrentPage int  `json:"current_page"`
+	PrevPage    *int `json:"prev_page"`
+	NextPage    *int `json:"next_page"`
+	TotalPages  int  `json:"total_pages"`
+	TotalCount  int  `json:"total_count"`
+	perPageNum  int
+}
+
+type Service interface {
+	Paginate(ctx context.Context, DBQuery bson.M, collection string, result interface{}, sortField string) (*Meta, error)
```
johnsudaar, line 46:
```suggestion
	Paginate(ctx context.Context, DBQuery bson.M, collection string, result interface{}, sortField string) (Meta, error)
```
Shouldn't we return the struct instead of a pointer to it ?

mongo/pagination/pagination.go
```diff
@@ -0,0 +1,172 @@
...
+	badRequestErr := NewBadRequestErrors()
+	pageErr := "Requested page"
+	perPageErr := "per_page"
+	var err error
+
+	// Options validation
+	if opts.PerPageDefault > opts.MaxPerPage ||
+		opts.PerPageDefault <= 0 ||
+		opts.MaxPerPage <= 0 {
+		return errors.New("Usage: MaxPerPage > PerPageDefault > 0")
```
johnsudaar, line 66:
```suggestion
		return errors.New("invalid pagination service configuration: MaxPerPage > PerPageDefault > 0")
```

mongo/pagination/pagination.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	// Parameter validation
+	if collection == "" {
+		return errors.New("collection must be set")
```
johnsudaar, line 70:
```suggestion
		return errors.New("invalid pagination service configuration: collection must be set")
```

mongo/pagination/pagination.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	if opts.PerPageQueryParams == "" {
+		opts.PerPageQueryParams = fmt.Sprintf("%d", opts.PerPageDefault)
+	}
+
+	// Request parameters validation
+	meta.CurrentPage, err = strconv.Atoi(opts.PageQueryParams)
+	if err != nil {
+		badRequestErr.Errors[pageErr] =
+			append(badRequestErr.Errors[pageErr], "fail to parse page parameter")
```
johnsudaar, line 85:
```suggestion
			append(badRequestErr.Errors[pageErr], fmt.Sprintf("%s is not a valid number", opts.PageQueryParams))
```

mongo/pagination/pagination.go
```diff
@@ -0,0 +1,172 @@
...
+	// Default values assignation
+	if opts.PageQueryParams == "" {
+		opts.PageQueryParams = "1"
+	}
+	if opts.PerPageQueryParams == "" {
+		opts.PerPageQueryParams = fmt.Sprintf("%d", opts.PerPageDefault)
+	}
+
+	// Request parameters validation
+	meta.CurrentPage, err = strconv.Atoi(opts.PageQueryParams)
```
johnsudaar, line 82: Should we check that opts.PageQueryParams is not an empty string first ? If the user does not specify a page we should fallback to the first one.

mongo/pagination/pagination.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		badRequestErr.Errors[pageErr] =
+			append(badRequestErr.Errors[pageErr], "fail to parse page parameter")
+	}
+	if meta.CurrentPage <= 0 {
+		badRequestErr.Errors[pageErr] =
+			append(badRequestErr.Errors[pageErr], "must be greater then 0")
+	}
+	meta.perPageNum, err = strconv.Atoi(opts.PerPageQueryParams)
```
johnsudaar, line 91: Same thing here.  if the PerPage is empty we should fallback to the default one.

mongo/pagination/pagination.go
```diff
@@ -0,0 +1,172 @@
...
+	TotalPages  int  `json:"total_pages"`
+	TotalCount  int  `json:"total_count"`
+	perPageNum  int
+}
+
+type Service interface {
+	Paginate(ctx context.Context, DBQuery bson.M, collection string, result interface{}, sortField string) (*Meta, error)
+}
+
+type ServiceOpts struct {
```
johnsudaar, line 49: I think that we should split this configutation in 2 different part:
* The PerPageDefault and MaxPerPage field should be set when initializing the PaginationService
* The PageQueryParams and PerPageQueryParams should be params passed to the Paginate method.
WDYT ?

mongo/pagination/pagination.go
curzolapierre: Yep, it is already check first, just a few lines above :)
We check here `PageQueryParams` and `PerPageQueryParams` if they are not empty, if they are we set a default value.

EtienneM (commented):
Can you add a README to explain how to use this?

mongo/pagination/README.md
```diff
@@ -0,0 +1,38 @@
...
+dbQuery := bson.M{"searched_field": "field_1"}
+
+meta, err := pageService.Paginate(
+    ctx,                    // A context (required)
+    "5",                    // Number of the requested page (can be empty, default 1)
+    "10",                   // Amount of items by page (can be empty)
+    dbQuery,                // Query which will be executed on the database (can be nil)
+    "RequestedCollection",  // Name of the collection (required)
+    &resultObject,          // The object that will contain the data (must be an array)
+    "-_id")                 // The field for the sort order (required)
```
EtienneM, line 34: When I read this code, I would prefer this function to take a structure as argument so that we understand the parameter. And the number should be of type `int`. But the mandatory args should remain in the outside this structure. E.g.

```go
type PaginateOpts struct {
  PageNumber int
  AmountItems int
  Query bson.M
  SortOrder string
}

Paginate(ctx context.Context, collection string, data interface{}, opts PaginateOpts)
```

WDYT?

EtienneM (approved):

johnsudaar (approved):

PR #168, opened 2021-01-25
Add pagination by cursor

curzolapierre:
Related to [SG-72]

[SG-72]: https://scalingo.atlassian.net/browse/SG-72

PR #169, opened 2021-01-28
Revert "Add pagination by cursor"

EtienneM:
After a thorough discussion with John and Pierre, we agreed this is not really a good idea. The Logs Archive will use the good ol' pagination (https://github.com/Scalingo/logs-archive-service/pull/35). 

Reverts Scalingo/go-utils#168


curzolapierre (approved):
LGTM

PR #194, opened 2021-05-05
fix: remove TODO

EtienneM:
I don't understand the TODO so I can't open an issue based on it


johnsudaar (approved):
If it's really an issue you'll see it soon enough ;)

PR #211, opened 2021-08-16
feature(nsqlbproducer) Add timeouts when publishing message

### LINKED ISSUE #210: [nsqlbproducer] Add a timeout to Publish and DeferredPublish
We should be able to configure a timeeout on Publish and DeferredPublish.
If a producer times out, it should go to the next one.

Related to [SRE-86]

[SRE-86]: https://scalingo.atlassian.net/browse/SRE-86

johnsudaar:
Fix #210
Related to [SRE-86]

[SRE-86]: https://scalingo.atlassian.net/browse/SRE-86

### DIFF
--- a/nsqlbproducer/lbproducer.go
+++ b/nsqlbproducer/lbproducer.go
@@ -21,7 +21,6 @@ const (
 )
 
 const (
-	PublishNoTimeout      = time.Duration(-1)
 	defaultPublishTimeout = 30 * time.Second
 )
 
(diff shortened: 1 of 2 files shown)


nsqlbproducer/lbproducer.go
```diff
@@ -20,6 +20,11 @@ const (
 	RandomStrategy
 )
 
+const (
+	PublishNoTimeout      = time.Duration(-1)
```
Soulou, line 24: Do we really want to let the possibility to have no timeout? I don't see any case we want that..

johnsudaar: i'm not sure but since it's not the default, i thought that it could be handy.
I can remove it if needed.

Soulou: I would remove it, it's a bad practice not having a timeout :)

Soulou (approved):

PR #218, opened 2021-10-21
[storage] Add Concurrency and PartSize support to S3 upload

Soulou:


storage/s3.go
```diff
@@ -7,17 +7,24 @@ import (
...
 )
 
 const (
 	NotFoundErrCode = "NotFound"
+	// DefaultPartSize 16 MB part size define the size in bytes of the parts
+	// uploaded in a multipart upload
+	DefaultPartSize = int64(16777216)
+	// DefaultUploadConcurrency defines that multipart upload will be done in
+	// parallel in 2 routines
+	DefaultUploadConcurrency = int(2)
```
EtienneM, line 27: Shouldn't they be private?
```suggestion
	// defaultPartSize 16 MB part size define the size in bytes of the parts
	// uploaded in a multipart upload
	defaultPartSize = int64(16777216)
	// defaultUploadConcurrency defines that multipart upload will be done in
	// parallel in 2 routines
	defaultUploadConcurrency = int(2)
```

Soulou: Usually I appreciate when Default values are public, as it's a constant it can't be changed and it might useful to be read from outside, it's not a risk to me

PR #262, opened 2022-02-25
feat(influx): query can now return the last element

curzolapierre:


influx/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## To be Released
 
+* Adding lastPoint to returns the last element
```
EtienneM, line 5:
```suggestion
* Adding LastPoint to returns the last element
```

influx/querybuilder.go
```diff
@@ -275,6 +275,11 @@ func (q Query) Fill(value string) Query {
 	return q
 }
 
+// LastPoint limits the query to return only the last element.
```
EtienneM, line 278: I think this comment should make it clear that it changes the `OrderByTime` of the query. I mean, if someone executes something like:

```go
influx.NewQuery().On(...).OrderByTime("ASC").LastPoint()
```

It can be a bit surprising.

influx/querybuilder.go
```diff
@@ -275,6 +275,11 @@ func (q Query) Fill(value string) Query {
...
+func (q Query) LastPoint() Query {
+	return q.OrderByTime("DESC").Limit(1)
```
EtienneM, line 280: Maybe add a constant for both `ASC` and `DESC` strings?

EtienneM (approved):
That's a good idea. Just a suggestion to improve the comment, otherwise LGTM

PR #379, opened 2022-09-16
feat(security): add a token manager

EtienneM:


Soulou (commented):
I'm not super fan in having the `TokenManager` with these 3 fields.

Personally I would have done a GenerateToken with ctx, payload, secrekey, validity.

Maybe there is a reason I haven't identified. I get why the `now` field for testing, definitely, but the rest I'm not that sure. Can you share your point of view?

But maybe there is a good reason I'm not detecting here.

EtienneM:
From my perspective it makes perfect sense that a `TokenManager` contains the configuration for all token generated and checked by this manager. The secret and the validity duration  fields are the manager configuration. We could decide that we want both these fields to be removed from the `TokenManager` structure, and add this information in a parameter of both the `GenerateToken` and `CheckToken` functions but I don't find it really elegant.

Soulou:
The question is how is used this structure.

Usually, is there one key for a lot of things, or each signature is authentified by another key (and in this case a Manager has to be created each time).

Indeed when asking this question, I think that most of the time, the key is common to multiple token, and in this case you're right.

EtienneM:
We currently use a similar token generator in git-repository-core (https://github.com/Scalingo/git-repository-core/blob/a85c214fa61e8ae89084b22effe46f33c91d2029/token/token.go) and scalingo-github-hook (https://github.com/Scalingo/scalingo-github-hook/blob/d25b37ab93e8a2458d0b00bc047a03bfb2da011d/services/public_archive_link_generator.go). We'll soon use it in the logs-service. 

I think that in these three cases, this architecture makes sense.

johnsudaar:
> Usually, is there one key for a lot of things, or each signature is authentified by another key (and in this case a Manager has to be created each time).

IMHO it sound safe to me to re-generate a manager every time we use another secret key. It could even be less error prone.

Plus the fact that we could handle the generator config in the dependency injection and nether worry about in the code seems like a plus (in controllers for examples). It sounds less error prone (copy/pasting would work with little adaptation needed).

Soulou (approved):

Soulou:
So it's all good to me Thanks @johnsudaar for your input

PR #393, opened 2022-10-12
feat(storage): Add Info method returns object information from header

curzolapierre:

`Info` method returns `ObjectNotFound` custom error in case of object not found

### DIFF
--- a/storage/backend.go
+++ b/storage/backend.go
@@ -3,15 +3,25 @@ package storage
 import (
 	"context"
 	"io"
+
+	"github.com/Scalingo/go-utils/storage/types"
 )
 
 // BackendMethod represents the name of a Method included in the Backend interface
 type BackendMethod string
 
+// ObjectNotFound is a custom error in case the object is not found
+type ObjectNotFound struct{}
+
+func (o ObjectNotFound) Error() string {
+	return "Object not found"
+}
+
 const (
 	GetMethod    BackendMethod = "Get"
 	UploadMethod BackendMethod = "Upload"
 	SizeMethod   BackendMethod = "Size"
+	InfoMethod   BackendMethod = "Info"
 	DeleteMethod BackendMethod = "Delete"
 )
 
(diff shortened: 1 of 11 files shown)


storage/backend.go
```diff
@@ -22,6 +22,7 @@ type Backend interface {
 	Upload(ctx context.Context, file io.Reader, path string) error
 	Size(ctx context.Context, path string) (int64, error)
 	Delete(ctx context.Context, path string) error
+	IsObjectExists(ctx context.Context, objectName string) (bool, error)
```
Soulou, line 25: What about having a method `Info` which return a custom error if it does not exist?

(I'm not a fan of the `IsObjectExists` function name --'

storage/s3.go
```diff
@@ -177,6 +179,37 @@ func (s *S3) Delete(ctx context.Context, path string) error {
...
+			if apiErr.ErrorCode() == NotFoundErrCode {
+				return types.Info{}, &ObjectNotFound{}
+			}
+		}
+		return types.Info{}, errors.Wrapf(err, "fail to HEAD object '%v'", path)
+	}
+
+	return types.Info{
+		ContentLength: res.ContentLength,
+		ContentType:   *res.ContentType,
```
Soulou, line 209: Normally both API would also have a MD5/SHA256, could you check that and include it?

Soulou (changes requested):

storage/s3.go
```diff
@@ -177,6 +179,38 @@ func (s *S3) Delete(ctx context.Context, path string) error {
...
+				return types.Info{}, ObjectNotFound{}
+			}
+		}
+		return types.Info{}, errors.Wrapf(err, "fail to HEAD object '%v'", path)
+	}
+
+	return types.Info{
+		ContentLength: res.ContentLength,
+		ContentType:   *res.ContentType,
+		Checksum:      *res.ChecksumSHA256,
```
Soulou, line 210: Please add the type of CHecksum too ;-) ChecksumSHA256, it may evolve over time.

THanks!

Soulou (changes requested):

curzolapierre:
Indeed it was a little more complicate than expected. Now it should be understood, thanks :+1: 

I put this here in case for me in the future: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html#checking-object-integrity-md5
About this ^
I guess logs-archives are not impacted by the 16MB limitation, but maybe we are in other case (e.g. backups).
If we want to check the integrity of each part of the upload we will have to support it specifically.

Soulou (approved):

PR #396, opened 2022-10-13
feat(mongo): contextify mongo.Session

EtienneM:
Without this change, our linter complains in all calling codes about the use of `context.Background` here. And I must say that I find it odd to pass the logger rather than the context here. @soulou Do you know if there is a reason about that?

This change should lead to a new major version of this lib

### DIFF
--- a/mongo/CHANGELOG.md
+++ b/mongo/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## To be Released
 
+* feat(mongo): contextify mongo.Session [#396](https://github.com/Scalingo/go-utils/pull/396)
 * build(deps): bump github.com/Scalingo/go-utils/logger from 1.1.1 to 1.2.0
 * build(deps): bump github.com/sirupsen/logrus from 1.8.1 to 1.9.0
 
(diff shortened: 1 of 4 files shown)


Soulou:
Yes there was a reason:

https://github.com/Scalingo/go-utils/pull/396/files#diff-11346246933b19ebec4e294c4e60d390fe1c10fca14b977a211e53c3a90387aaR26

The fact we're creating an initial connection and cloning it. It means that only the initial Session well have a context.

Once this context expires this session will be closed, and Cloning it won't be possible anymore, leading to errors everywhere.

Does it create a problem currently in any project?

Soulou (commented):
See my comment

EtienneM:
OK interesting, thanks for clarifying this. As stated in my PR description, the problem comes from our linter [`contextcheck`](https://github.com/kkHAIKE/contextcheck). It does not really like that we don't forward the context and create a new one with `context.Background` in this lib.

I think I will close this PR, and open a new one with more documentation:
- explaining why we take a logger and not a context in param.
- explaining why the linter will probably fail and how to punctually skip the check (`// nolint: contextcheck`)

Soulou:
Ok great, there are things to improve this lib, but I think it's more tricky than just adding this context

PR #425, opened 2022-11-07
feat(storage): `Delete` may return `ObjectNotFound`

EtienneM:

### DIFF
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -178,6 +178,12 @@ func (s *S3) Delete(ctx context.Context, path string) error {
 	input := &s3.DeleteObjectInput{Bucket: &s.cfg.Bucket, Key: &path}
 	_, err := s.s3client.DeleteObject(ctx, input)
 	if err != nil {
+		var apiErr smithy.APIError
+		if stderrors.As(err, &apiErr) {
+			if apiErr.ErrorCode() == NotFoundErrCode {
+				return ObjectNotFound{}
+			}
+		}
 		return errors.Wrapf(err, "fail to delete object %v", path)
 	}
 
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -128,7 +128,7 @@ func (s *S3) Get(ctx context.Context, path string) (io.ReadCloser, error) {
 	}
 	out, err := s.s3client.GetObject(ctx, input)
 	if err != nil {
-		return nil, errors.Wrapf(err, "fail to get object %v", path)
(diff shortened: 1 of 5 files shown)


storage/swift.go
```diff
@@ -116,7 +119,7 @@ func (s *Swift) Info(ctx context.Context, path string) (types.Info, error) {
 	info, _, err := s.conn.Object(ctx, s.cfg.Container, path)
 	if err != nil {
 		if err.Error() == swift.ObjectNotFound.Error() {
-			return types.Info{}, &ObjectNotFound{}
+			return types.Info{}, ObjectNotFound{}
```
EtienneM, line 122: The `Info` command in s3.go returns `ObjectNotFound{}` as a value (and not a pointer). Let's uniformize as it's creating problems in the calling codes

storage/s3.go
```diff
@@ -128,7 +128,7 @@ func (s *S3) Get(ctx context.Context, path string) (io.ReadCloser, error) {
 	}
 	out, err := s.s3client.GetObject(ctx, input)
 	if err != nil {
-		return nil, errors.Wrapf(err, "fail to get object %v", path)
+		return nil, errors.Wrapf(err, "fail to get S3 object %v", path)
```
EtienneM, line 131: I found it unclear in the Rollbar errors that the storage backend is not in the error message. Hence I wanted to add it. It's a separated commit so if you don't like it I can revert easily.

storage/s3.go
```diff
@@ -178,7 +178,13 @@ func (s *S3) Delete(ctx context.Context, path string) error {
 	input := &s3.DeleteObjectInput{Bucket: &s.cfg.Bucket, Key: &path}
 	_, err := s.s3client.DeleteObject(ctx, input)
 	if err != nil {
-		return errors.Wrapf(err, "fail to delete object %v", path)
+		var apiErr smithy.APIError
+		if stderrors.As(err, &apiErr) {
+			if apiErr.ErrorCode() == NotFoundErrCode {
+				return ObjectNotFound{}
```
Soulou, line 184: Would you mind adding the `Path` in the `ObjectNotFound`

Why not having `ObjectNotFoundError` ? :D

EtienneM: > Why not having `ObjectNotFoundError` ? :D

Because it means a change in the API, so a new major version... 😓 

> Would you mind adding the `Path` in the `ObjectNotFound`

Good idea, on it

Soulou: Oooh, no problem for the first part

PR #431, opened 2022-11-23
fix: privatify `ErrgoRoot`

EtienneM:
We shouldn't use it anymore. We should use `RootCause` in place.

We break the API so it should lead to a new major version but it has been decided with @Soulou that the change is really small and it's not necessary.


Soulou (commented):
Could you update, your comment? I think we should do a major version, we're breaking the API

Soulou (approved):

PR #445, opened 2022-12-14
feat(errors): `RootCtx` retrieves the deepest context from wrapped errors

### LINKED ISSUE #444: [ErrCtx] Be able to retrieve the deepest context

curzolapierre:

Fix #444


errors/errctx.go
```diff
@@ -48,3 +49,60 @@ func Wrapf(ctx context.Context, err error, format string, args ...interface{}) e
...
+
+	var lastCtx context.Context
+
+	type causer interface {
+		Cause() error
+	}
+
+	// Unwrap each error to get the deepest context
+	for err != nil {
+		log.Debug("Unwrapping error type '%T'\n", err)
```
Soulou, line 66: I'm really not super fan of adding custom debug logging (impacted by environment) in a lib like t his, tests should ensure it works. These `log.Debug` are to be removed for me.

errors/errctx_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/errgo.v1"
+
+	"github.com/Scalingo/go-utils/logger"
+)
+
+func TestErrCtx_GetDeepestContext(t *testing.T) {
+	t.Run("It should return a nil context if the error is not wrapped with ErrCtx", func(t *testing.T) {
```
Soulou, line 17: Don't we want to never return nil (to prevent having to check for `nil` and return a default context? (`context.Background()`) ?

errors/errctx_test.go
```diff
@@ -0,0 +1,155 @@
...
+		deepestCtx := GetDeepestContext(err)
+
+		// Then
+		assert.Nil(t, deepestCtx)
+		assert.Contains(t, err.Error(), "wrapping error in func2")
+	})
+
+	t.Run("It should get the deepest context and contains fields from multiple wrapped error", func(t *testing.T) {
+		// Given
+		ctx, _ := logger.WithFieldToCtx(context.Background(), "field0", "value0")
```
Soulou, line 32: Could you not use the `log` libs, and use `context.Withvalue` ? from the `context` package? (it would prevent casting below)

errors/errctx.go
```diff
@@ -48,3 +49,60 @@ func Wrapf(ctx context.Context, err error, format string, args ...interface{}) e
 func Errorf(ctx context.Context, format string, args ...interface{}) error {
 	return ErrCtx{ctx: ctx, err: errors.Errorf(format, args...)}
 }
+
+// GetDeepestContext unwrap all wrapped errors from err to get the deepest context
+// from ErrCtx errors. If there is no wrapped ErrCtx, GetDeepestContext returns nil.
+func GetDeepestContext(err error) context.Context {
```
Soulou, line 55: Small question about the naming, why not using `RootCtx` to match the other method names?

Soulou (changes requested):

errors/errctx_test.go
curzolapierre: Finally I'm wonder about this one.
In case we are not sure about the error type get could still call `RootCtx`. If no `ErrCtx` found we probably want to notice it to avoid the usage of a `context.Background()`.
If `RootCtx` returns nil, it is clear that there is no `ErrCtx` error wrapped in the given err.

WDYT?

errors/errctx_test.go
```diff
@@ -0,0 +1,153 @@
...
+	stdErrors "errors"
+	"testing"
+
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/errgo.v1"
+)
+
+func TestErrCtx_RootCtx(t *testing.T) {
+	t.Run("It should return a default context if the error is not wrapped with ErrCtx", func(t *testing.T) {
```
curzolapierre, line 14: IMO we should return nil in case of no `ErrCtx` wrapped in the given err

errors/errctx_test.go
Soulou: You're right, but it makes the usage trickier since it's not sure the user will now the `ctx` can be `nil`, it's really not a habit in Go

curzolapierre: True, what about returning an err?

curzolapierre: I've added a "not found" error in addition

EtienneM (commented):
I don't feel confident reviewing this code. I didn't follow what has been done and I feel a bit lost. I would trust @Soulou on this review.

errors/errctx_test.go
Soulou: I'm adding @johnsudaar to the loop since I need an external point of view on this API, I don't feel super super confortable with the error in `RootCtx`

curzolapierre: FYI, with the current implementation, it could be used like this: https://github.com/Scalingo/go-handlers/pull/57/files#diff-683abe1277a9083805e9aeb9965a862d1d5e422c500ed7d9a41f61e75f306f9aR42-R44
But I share your feel, I'm not comfortable with the error too.

johnsudaar: I'm not found of this error either.

Maybe another implementation could be changing the method signature ?
```golang
func RootCtx(ctx context.Context, err error) (context.Context)
```

And send the context passed as argument if the context is not found in the error.

```golang
if err != nil {
  logger.Get(errctx.RootCtx(ctx, err)).WithError(err).Error("Something bad happened)
}
```

Soulou: Ok, I like @johnsudaar 's idea, thanks a lot for the help

Maybe in this case `RootCtx` is not the right name, or `err` should be first, since `ctx` is the "fallback" context.

```
errctx.RootCtxOrFallback(err, ctx)
```

(Simple proposal

Soulou (commented):
See above comment

errors/errctx_test.go
Soulou: Yep but it's annoying to have linter errors to ignore haha, it won't be the last because of this

Soulou (approved):

PR #458, opened 2023-01-05
fix(errctx): remove NoteMask, `Notef` should be use instead

### LINKED ISSUE #457: [ErrCtx] NoteMask should take a pass function in parameter according to errgo

curzolapierre:
Fix #457

### DIFF
--- a/errors/CHANGELOG.md
+++ b/errors/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## To be Released
 
+* fix: remove `NoteMask`. `Notef` should be use instead.
+
 ## v2.2.0
 
 * feat: add UnwrapError to unwrap one error.
--- a/errors/errctx.go
+++ b/errors/errctx.go
@@ -28,10 +28,6 @@ func Newf(ctx context.Context, format string, args ...interface{}) error {
 	return ErrCtx{ctx: ctx, err: errgo.Newf(format, args...)}
 }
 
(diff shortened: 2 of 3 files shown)


errors/errctx.go
```diff
@@ -28,8 +28,8 @@ func Newf(ctx context.Context, format string, args ...interface{}) error {
 	return ErrCtx{ctx: ctx, err: errgo.Newf(format, args...)}
 }
 
-func NoteMask(ctx context.Context, err error, message string) error {
-	return ErrCtx{ctx: ctx, err: errgo.NoteMask(err, message)}
+func NoteMask(ctx context.Context, err error, message string, pass ...func(error) bool) error {
+	return ErrCtx{ctx: ctx, err: errgo.NoteMask(err, message, pass...)}
 }
```
EtienneM, line 33: Sorry I didn't know about this function but it should not be implemented. We don't use and `*Mask*` function since a loooong time. Any call to a Mask function should be converted to a `Notef` call.
```suggestion
```

errors/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## To be Released
 
+* fix: `NoteMask` is taking a pass function as parameter
```
EtienneM, line 5: The changelog does not look that good anymore :p

EtienneM (approved):

PR #460, opened 2023-01-10
Bump nsqconsumer to v1.2.0: better error management

Soulou:


EtienneM (changes requested):
You also need to change the version number in the README as specified in the procedure here: https://github.com/Scalingo/go-utils#release-a-new-version-of-a-module

EtienneM (approved):

PR #583, opened 2023-04-05
feat(Is/As) Match API from standard library and leverage it to check for error value of type

### LINKED ISSUE #582: [errors] Package should leverage standard library errors.Is and errors.As methods since the ecosystem is going toward this direction
As a follow up of some discussions of our workshop

Soulou:
I'm proposing to deprecate `IsRootCause(err, errWithExpectedType)` and `if(errors.RootCause(err) == something` to use `errors.As(err, errWithExpectedType)` and `errors.Is(err, something)`, to match idiomatic Go practices

Fixes #582

### DIFF
--- a/errors/CHANGELOG.md
+++ b/errors/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## To be Released
 
+* feat: add `Is` and `As` to match standard library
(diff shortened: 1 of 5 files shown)


Soulou:
I'll update our internal doc and make an announce of course

EtienneM (approved):

PR #585, opened 2023-04-06
feat(v3) Drop Notef, drop Cause(), drop errgo entirely

Soulou:
Merge after #584


EtienneM (approved):
I liked the idea!

leo-scalingo:
Yes but not that mandatory, closing for now

PR #586, opened 2023-04-13
[errors] Bump v2.3.0

Soulou:

### DIFF
--- a/errors/CHANGELOG.md
+++ b/errors/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## To be Released
 
+## v2.3.0
+
 * feat: add `Is` and `As` to match standard library
 * fix: remove `NoteMask`. `Notef` should be use instead.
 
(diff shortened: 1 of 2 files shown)


errors/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## To be Released
 
+## v2.3.0
+
 * feat: add `Is` and `As` to match standard library
 * fix: remove `NoteMask`. `Notef` should be use instead.
```
EtienneM, line 8: Don't we consider that this item should lead to a new major version?

EtienneM (approved):
Just a question but already approving this PR.

errors/CHANGELOG.md
Soulou: Hmm indeed I didn't check I was breaking something here.

Actually I would want to make a v3 with the removal of errgo, it will come.

PR #707, opened 2023-07-07
feat: implement go-handlers 400 errors and return 200 when the page is empty

### LINKED ISSUE #706: [mongo] Pagination service should not throw error but returns empty list instead
According to the proposition here: https://scalingo.slack.com/archives/C01E42AT2G6/p1688550147298259

We probably want to stop throwing badRequest error from pagination service but just return an empty list..

We need to think about the impacted services and their error management.

SCedricThomas:

fix #706


mongo/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## To be Released
 
+feat(pagination): implement go-handlers 400 errors and return 200 when the page is empty
```
curzolapierre, line 5: Nitpick: the sentence is not really understandable, maybe something like:
```suggestion
feat(pagination): replace BadRequestError by the one from go-handlers to return HTTP 400.
feat(pagination): return HTTP 200 when the page is empty instead of BadRequestError
```
would be better, WDYT?

The idea is: it is a breaking change, we need to be explicit about it.

mongo/pagination/pagination_test.go
```diff
@@ -65,12 +65,51 @@ func TestPaginationPaginate(t *testing.T) {
...
+					PrevPage:    &prevPage,
+					NextPage:    nil,
+					TotalPages:  2,
+					TotalCount:  4,
+					perPageNum:  2,
+				}
+			},
+		},
+		{
+			Name: "It should return an error when the requested AmountItems is superior to MaxPerPage",
```
curzolapierre, line 84:
```suggestion
			Name: "It should return an error when the requested AmountItems is greater than MaxPerPage",
```

mongo/pagination/pagination_test.go
```diff
@@ -65,12 +65,51 @@ func TestPaginationPaginate(t *testing.T) {
...
+			PaginationOpts: &ServiceOpts{
+				PerPageDefault: 2,
+				MaxPerPage:     2,
+			},
+			PageNumber:  0,
+			AmountItems: 3,
+			Error:       "* per_page → must be lower or equal to 2",
+		},
+		{
+			Name: "It should return an error when the requested AmountItems is inferior to 0",
```
curzolapierre, line 98:
```suggestion
			Name: "It should return an error when the requested AmountItems is lower than 0",
```

mongo/pagination/pagination_test.go
```diff
@@ -65,12 +65,51 @@ func TestPaginationPaginate(t *testing.T) {
...
+				PerPageDefault: 2,
+				MaxPerPage:     2,
+			},
+			PageNumber:  0,
+			AmountItems: -1,
+			Error:       "* per_page → must be greater than 0",
 		},
 		{
-			Name: "It should return an error with a request item per page superior then max per page",
+			Name: "It should return an error when the requested item per page is superior to max per page",
```
curzolapierre, line 112:
```suggestion
			Name: "It should return an error when the requested item per page is greater than max per page",
```

curzolapierre (changes requested):
There are others `then` that should be `than`. Could you replace them please?

Otherwise, only some nitpicks, thanks for handling this :+1:

mongo/pagination/pagination_test.go
```diff
@@ -329,6 +368,7 @@ func TestPaginationPaginate(t *testing.T) {
 			meta, err := run.PaginationOpts.Paginate(context.Background(),
 				dummyCollection, &results, paginateOpts)
 			if run.Error != "" {
+				require.NotNil(t, err)
```
curzolapierre, line 371: Suggestion: 
```suggestion
				require.Error(t, err)
```

curzolapierre (changes requested):
There is still `then` in the pagination module, e.g. https://github.com/Scalingo/go-utils/pull/707/files#diff-e90d6bda755cb9479c9f732f823ac4aa98ebff14bc0bb7c1607d0d9bb5418cd6R74

curzolapierre (approved):
:+1:

PR #858, opened 2024-03-28
feat(storage): add allow list check on S3 creation

### LINKED ISSUE #804: [S3] - add an allowlist of domain names for uploading files to storage
To enforce security, we would like to add an allowlist to make sure we upload our files to a trusted object storage.

sihamais:

[NCP-555]
Fix #804

[NCP-555]: https://scalingo.atlassian.net/browse/NCP-555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -5,6 +5,8 @@ import (
 	stderrors "errors"
 	"fmt"
 	"io"
+	"os"
+	"slices"
 	"strings"
 	"time"
 
@@ -302,12 +304,14 @@ func s3Config(cfg S3Config) aws.Config {
 		Credentials: credentials,
 	}
 	if cfg.Endpoint != "" {
-		config.EndpointResolverWithOptions = aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
-			return aws.Endpoint{
-				URL:           "https://" + cfg.Endpoint,
-				SigningRegion: cfg.Region,
-			}, nil
-		})
+		if endpointIsAllowed(cfg.Endpoint) {
+			config.EndpointResolverWithOptions = aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
+				return aws.Endpoint{
+					URL:           "https://" + cfg.Endpoint,
+					SigningRegion: cfg.Region,
+				}, nil
+			})
+		}
 	}
 
 	return config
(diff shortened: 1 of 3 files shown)


storage/s3.go
```diff
@@ -62,29 +62,43 @@ type S3 struct {
...
+			if strings.HasPrefix(s3.cfg.Endpoint, url) {
+				return nil
+			}
+		}
+		return errors.New("endpoint is not in allow list")
 	})
 }
 
-func NewS3(cfg S3Config, opts ...s3Opt) *S3 {
+func NewS3(cfg S3Config, opts ...S3Opt) (*S3, error) {
```
curzolapierre, line 101: Issue: Usually we don't want `New` functions to return error. It should only instantiate the object.
In addition, this will break retro-compatibility.

Suggestion: From https://github.com/Scalingo/go-utils/issues/804#issuecomment-2024912293 suggestion is to log something in run time.
But if you really prefer to throw an error, I'd see either:
- return it in runtime (e.g. `Get`, `Upload`, ...) but not very practical in fact
- create a dedicated helper to validate the list. But it need to add this check everywhere.
- keep it like this. In that case take look at the current implementations in our code base

storage/s3.go
```diff
@@ -62,29 +62,43 @@ type S3 struct {
 	partSize          int64
 }
 
-type s3Opt func(s3 *S3)
+type S3Opt func(s3 *S3) error
```
curzolapierre, line 65: Question: what is the reason for exporting this signature?

IMO it should stay private in order to only provide `With*` functions here. Or am I missing use cases?

storage/s3.go
sihamais: I changed the way it's handled. WDYT?

curzolapierre: Sorry for the delay and the change of mind, as we are in a context of configuration issue we should get an error as soon as possible. Hence the need to use an error in constructor (so your proposal in this commit: https://github.com/Scalingo/go-utils/compare/d599d4eee09fceb8c27e60c30dff5fa7b5a238dc...fddf2c502859a62b128e658aa53fd0287a7f6ad5) :+1:

From our implementation in code base (e.g. [on database-core](https://github.com/Scalingo/appsdeck-database/blob/e3964de87b54eb736b20d9b74db1a9fbe0d79047/controller/backup/storage/s3/s3.go#L30-L66)) we should make some change to fit the logic, i.e add a `BackendGetter` struct on boot of the related service.

Note: this will imply to make a major update of the package.

For the record: https://scalingo.slack.com/archives/CRC7L0XM2/p1714063053317099?thread_ts=1714058866.915739&cid=CRC7L0XM2

storage/s3.go
```diff
@@ -302,12 +304,15 @@ func s3Config(cfg S3Config) aws.Config {
 		Credentials: credentials,
 	}
 	if cfg.Endpoint != "" {
-		config.EndpointResolverWithOptions = aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
-			return aws.Endpoint{
-				URL:           "https://" + cfg.Endpoint,
-				SigningRegion: cfg.Region,
-			}, nil
-		})
+		if endpointIsAllowed(cfg.Endpoint) {
```
curzolapierre, line 307: Suggestion(wording):
```suggestion
		if isEndpointAllowed(cfg.Endpoint) {
```

yanjost:
@curzolapierre when you say "major update", what does this imply ? We're trying to keep a small scope in this change...

yanjost:
As discussed with @leo-scalingo InfoSec will add the checks in the library and let IST update the projects during normal maintenance cycles

Issue #31, opened 2018-06-06
Query builder should be completly stateless


EtienneM:
Which is not the case for the condition. See https://github.com/Scalingo/go-utils/pull/28#issuecomment-394722403

PR #909, opened 2024-06-21
Fix/31/replace grace with tableflip

sc-zenokerr:
Update graceful to use tableflip

Update graceful test helpers, to allow asynchronous timed checks
Related to acadock issue: [#31](https://github.com/Scalingo/acadock-monitoring/issues/31) 


graceful/service.go
```diff
@@ -9,15 +9,16 @@ import (
...
-	"gopkg.in/errgo.v1"
+	"github.com/Scalingo/go-utils/errors/v2"
+
+	"github.com/cloudflare/tableflip"
 
 	"github.com/Scalingo/go-utils/logger"
 )
 
-type service struct {
+type Service struct {
```
sc-zenokerr, line 19: I've exported the "Service" struct. I'm not sure if you will agree with this, but it is generally better practice as it maintains type safety. 
Another option would be for NewService to return an interface, we could then keep the service struct private.
I did not think this would be a dangerous change, as there are no exported fields within the struct.

graceful/service_test.go
sc-zenokerr, line 1: I've updated this file for a couple of reasons.

1. The original task issue 31 in `acadock-monitoring` was to replace grace with tableflip.
   * I wrote tests for this code (following the pattern of the tests here), in doing so, I made some improvements to the tests, and it would have been a shame to waste them.
   * Within the original test, there were process checks that were intended to be async, but were running synchronously. The new test code explicitly allows async and sync checks.
   * The test code i have written is "unexported", but it is however decoupled from the tests themselves, so we would be able to lift this out to a library if ever we needed to.

graceful/service.go
leo-scalingo: Actually I'm good with it, returning a private entity from a public package is not a good practice, indeed a public struct or an interface would do. Considering this a good improvement :+1:

graceful/service.go
```diff
@@ -76,69 +76,99 @@ func (s *service) ListenAndServeTLS(ctx context.Context, proto string, addr stri
...
-func (s *service) listenAndServe(ctx context.Context, proto string, addr string, server *http.Server) error {
+func (s *Service) listenAndServe(ctx context.Context, _ string, addr string, server *http.Server) error {
+	log := logger.Get(ctx)
+
 	if s.pidFile != "" {
 		pid := os.Getpid()
 		err := os.WriteFile(s.pidFile, []byte(fmt.Sprintf("%d\n", pid)), 0600)
 		if err != nil {
-			return errgo.Notef(err, "fail to write PID file")
+			return errors.Wrap(ctx, err, "fail to write PID file")
```
leo-scalingo, line 94: Good reflex are already coming :-D

graceful/service.go
```diff
@@ -76,69 +76,99 @@ func (s *service) ListenAndServeTLS(ctx context.Context, proto string, addr stri
...
 
-	ld, err := s.graceful.Listen(proto, addr)
+	// Use tableflip to handle graceful restart requests
+	upg, err := tableflip.New(tableflip.Options{
+		UpgradeTimeout: s.reloadWaitDuration,
+		PIDFile:        s.pidFile,
+	})
 	if err != nil {
-		return errgo.Notef(err, "fail to get listener")
+		log.Fatalln(err)
```
leo-scalingo, line 104: Why changing from returning an error, to crashing the software?

graceful/service.go
```diff
@@ -76,69 +76,99 @@ func (s *service) ListenAndServeTLS(ctx context.Context, proto string, addr stri
...
 	if err != nil {
-		return errgo.Notef(err, "fail to shutdown server")
+		return errors.Wrapf(ctx, err, "fail to shutdown server")
 	}
-	return s.waitStopped()
+
+	// Wait for connections to drain.
+	err = server.Shutdown(ctx)
+	if err != nil {
+		log.Println("server shutdown:", err)
```
leo-scalingo, line 154: With our logger, we're never using `Println`, we use `Debug` / `Debugf` - `Info` / `Infof` - `Error` / `Errorf` and that's it. 

No `Warn` because it's ambiguous (is it  a problem or not? If yes, it's an error to trigger a Rollbar notification, if not, it's just information, since no one is reviewing `Warn`

`Println` is quite similar to `Info` usually, but less explicit, so we really prefer `Info`

graceful/service.go
```diff
@@ -76,69 +76,99 @@ func (s *service) ListenAndServeTLS(ctx context.Context, proto string, addr stri
...
+		ln = tls.NewListener(ln, server.TLSConfig)
 	}
 
 	s.httpServer = server
 
-	go s.setupSignals(ctx)
+	go func() {
+		err := server.Serve(ln)
+		if !errors.Is(err, http.ErrServerClosed) {
+			log.Println("HTTP server:", err)
```
leo-scalingo, line 129: Like below `Println` → `Info` 

So this error is "normal" right?  It's an info to tells the server has shutdown normally?

graceful/service.go
```diff
@@ -76,69 +76,99 @@ func (s *service) ListenAndServeTLS(ctx context.Context, proto string, addr stri
...
+	s.graceful = upg
+
+	// setup the signal handling
+	go s.setupSignals(ctx)
+
+	// Listen must be called before Ready
+	ln, err := upg.Listen("tcp", addr)
+	if err != nil {
+		upg.Stop()
+		log.Fatalln("Can't listen:", err)
```
leo-scalingo, line 115: Same as above

leo-scalingo (changes requested):

graceful/service.go
sc-zenokerr: This is an error on my part. Thanks for picking it up.

graceful/service.go
sc-zenokerr: I have changed this to log.Error.

 I think it is better to log an error as we should be made aware if this fails regularly.

graceful/service_signals.go
```diff
@@ -12,40 +11,23 @@ import (
...
-			ctx, cancel := context.WithTimeout(ctx, s.reloadWaitDuration)
-			defer cancel()
-			err = s.shutdown(ctx)
+			err := s.graceful.Upgrade()
 			if err != nil {
-				s.stopped <- fmt.Errorf("fail to shutdown service: %v", err)
-			} else {
-				close(s.stopped)
-				return
+				log.Println("fail to start new service:", err)
```
leo-scalingo, line 30:
```suggestion
				log.WithError(err).Error("fail to start new service")
```

graceful/service_signals.go
```diff
@@ -12,40 +11,23 @@ import (
...
-			ctx, cancel := context.WithTimeout(ctx, s.reloadWaitDuration)
-			defer cancel()
-			err = s.shutdown(ctx)
+			err := s.graceful.Upgrade()
 			if err != nil {
-				s.stopped <- fmt.Errorf("fail to shutdown service: %v", err)
-			} else {
-				close(s.stopped)
-				return
+				log.WithError(err).Error("fail to start new service")
```
leo-scalingo, line 30: I'm wondering, why did you change the behavior here? (from sending to the `stopped` channel vs logging the error? It's not really clear to me

sc-zenokerr: Thank you for pointing this out, I have looked at it again, and will try to explain.

* Previously, there were 2 separate errors; one for `StartProcess` was logged; and one as a result of `shutdown` (of parent process) timing out (sent to the `stopped` channel).
* Now, there is only one error `Upgrade` which is logged. The upgrade does *not* include the termination of the existing process.

Now, the shutdown timeout is handled in file signals.go:142 . If an error occurs or the shutdown times out, the error is returned. This is what the `stopped` channel was used for. This functionality is tested in the test `TestService_Shutdown_WithTimeout`

In fact I should have removed the `stopped` channel, and will push another commit that does so.

This is an important library with regards to the service, so if you think have more comments, please let me know.

leo-scalingo: Your comment suits me, and indeed in this case the channel should be removed

leo-scalingo (approved):

PR #928, opened 2024-07-10
Add --output-path flag (OUTPUT_PATH environment variable)

sc-zenokerr:


EtienneM (commented):
Are you sure we need this? We have been using gomock_generator for years without needing it

sc-zenokerr:
For me, When I am working on a project, I checking out a copy of the project to a different folder. Thus leaving the gogo2 updated version of the project untouched.

For example, my WIP version of a project would be in `~/GolandProjects/appsdeck-database` rather than `~/go/src/github.com/Scalingo/appsdeck-database`

Currently the gomock_generator does not allow me to target my development copy. The addition of the `output-path` option allows me to generate the mocks in the folder I am developing in.

This option therefore provides some flexibility for the local development environment.

sc-zenokerr:
This PR is not ready... I'll come back with another one later.

PR #930, opened 2024-07-11
gomock_generator. Add --source-path option

sc-zenokerr:

Add a --source-path option that allows for the user to specify the location of their codebase.

e.g. gomocks_generator --source-path <path_to_codebase>

### DIFF
--- a/gomock_generator/gomockgenerator/generate.go
+++ b/gomock_generator/gomockgenerator/generate.go
@@ -28,6 +28,8 @@ type GenerationConfiguration struct {
 	ConcurrentGoroutines int
 	// NoGoMod by default we'll consider go modules is enabled, mockgen will be called with -mod=mod to read interfaces in modules instead of default GOPATH
 	NoGoMod bool
+	// SourcePath override the output base path of the mocks
+	SourcePath string
 }
 
 // MocksConfiguration contains the configuration of the mocks to generate.
@@ -67,7 +69,12 @@ func GenerateMocks(ctx context.Context, gcfg GenerationConfiguration, mocksCfg M
 	if mocksCfg.BaseDirectory == "" {
 		mocksCfg.BaseDirectory = mocksCfg.BasePackage
 	}
-	err = os.Chdir(path.Join(os.Getenv("GOPATH"), "src", mocksCfg.BaseDirectory))
+
+	mocksPath := path.Join(path.Join(os.Getenv("GOPATH"), "src"), mocksCfg.BaseDirectory)
+	if gcfg.SourcePath != "" {
+		mocksPath = gcfg.SourcePath
+	}
+	err = os.Chdir(mocksPath)
 	if err != nil {
 		return errors.Wrap(err, "fail to move to base package directory")
 	}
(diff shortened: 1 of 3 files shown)


EtienneM (commented):
I'm not sure I understand the need for this option. Is it to be able to execute `gomock_generator` from another folder than the service root folder?

sc-zenokerr:
Hi @EtienneM .

The reason is that the local version of the codebase might not be in the default location.

Prior to this change, a codebase such as Scalingo/go-utils would be located $GOPATH/src/github.com/Scalingo/swan-agent

gomock_generator REQUIRES the codebase to be located in the default location. 

The --source-path option allows us to locate the codebase somewhere other than the default location.

gomock_generator/gomockgenerator/generate.go
```diff
@@ -28,6 +28,8 @@ type GenerationConfiguration struct {
 	ConcurrentGoroutines int
 	// NoGoMod by default we'll consider go modules is enabled, mockgen will be called with -mod=mod to read interfaces in modules instead of default GOPATH
 	NoGoMod bool
+	// SourcePath override the output base path of the mocks
+	SourcePath string
```
EtienneM, line 32: issue: I'm not sure I understand the choice of words here for the variable name. It mentions `Source` but it seems to be the location of what this program outputs. What do you think about `MocksPath` or `MocksRootPath`?

gomock_generator/gomockgenerator/generate.go
```diff
@@ -79,7 +86,10 @@ func GenerateMocks(ctx context.Context, gcfg GenerationConfiguration, mocksCfg M
 	}).Infof("Generating %v mocks", len(mocksCfg.Mocks))
 
 	var mockSigs map[string]string
-	mockSigsPath := path.Join(os.Getenv("GOPATH"), "src", mocksCfg.BaseDirectory, gcfg.SignaturesFilename)
+	mockSigsPath := path.Join(gcfg.SourcePath, mocksCfg.BaseDirectory, gcfg.SignaturesFilename)
+	if gcfg.SourcePath != "" {
+		mockSigsPath = path.Join(gcfg.SourcePath, gcfg.SignaturesFilename)
+	}
```
EtienneM, line 92: issue: from what I understand, it shouldn't be needed at this point to still have a condition. `mocksPath` contains the value we need.

```suggestion
	mockSigsPath := path.Join(mocksPath, gcfg.SignaturesFilename)
```

gomock_generator/gomockgenerator/generate.go
```diff
@@ -67,7 +69,12 @@ func GenerateMocks(ctx context.Context, gcfg GenerationConfiguration, mocksCfg M
 	if mocksCfg.BaseDirectory == "" {
 		mocksCfg.BaseDirectory = mocksCfg.BasePackage
 	}
-	err = os.Chdir(path.Join(os.Getenv("GOPATH"), "src", mocksCfg.BaseDirectory))
+
+	mocksPath := path.Join(path.Join(os.Getenv("GOPATH"), "src"), mocksCfg.BaseDirectory)
```
EtienneM, line 73: issue: `path.Join` receives variadic arguments
```suggestion
	mocksPath := path.Join(os.Getenv("GOPATH"), "src", mocksCfg.BaseDirectory)
```

gomock_generator/gomockgenerator/generate.go
```diff
@@ -67,7 +69,12 @@ func GenerateMocks(ctx context.Context, gcfg GenerationConfiguration, mocksCfg M
...
+	if gcfg.SourcePath != "" {
+		mocksPath = gcfg.SourcePath
+	}
```
EtienneM, line 76: issue: IMO the check with the argument `SourcePath` should only be done once. The fact that you need to do the check over and over again is a sign that there is a lack of optimization somewhere. Doing this check here, at the beginning of the mocks generation is ok. Doing it multiple times makes me feel like something is wrong. Can you have a look to optimize this?

EtienneM (changes requested):

gomock_generator/gomockgenerator/generate.go
sc-zenokerr: I'm not precious about the naming... but it is the location of your codebase. MocksRootPath works. I'll update it.

gomock_generator/gomockgenerator/generate.go
sc-zenokerr: Yep. indeed.

EtienneM:
@sc-zenokerr long time with no update. Is it still relevant?

PR #995, opened 2024-11-04
feat(logger): add WithRedactedFields configuration option

sc-zenokerr:
Adds a "WithRedactedFields" configuration option that allows for log fields to be redacted using regular expressions.

Part of [STORY-1053]


logger/redacting_formatter.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"regexp"
+
+	"github.com/sirupsen/logrus"
+)
+
+type RedactionOption struct {
+	Field       string
+	Regexp      *regexp.Regexp
```
sc-david-voisin, line 11: Do we need a regexp here ?  Wouldn't a string be enough ?
I am afraid on performances with huge log volumes.

logger/redacting_formatter.go
```diff
@@ -0,0 +1,48 @@
...
+	ReplaceWith string
+}
+
+type RedactingFormatter struct {
+	logrus.Formatter
+	fields []*RedactionOption
+}
+
+func (f *RedactingFormatter) Format(entry *logrus.Entry) ([]byte, error) {
```
sc-david-voisin, line 20: Even if covered in `logger_test.go`, I would add a dedicated test for this method, and cover more cases (as per the "[REDACTED]" replacement).
What do you think ?

logger/logger_test.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDefault(t *testing.T) {
+	logger := Default()
+	assert.NotNil(t, logger)
+	assert.Equal(t, logrus.InfoLevel, logger.(*logrus.Logger).Level)
+}
+
```
sc-david-voisin, line 18: praise: GG for adding the missing tests !

sc-david-voisin (commented):
praise: GG for adding the missing tests !

logger/config_test.go
```diff
@@ -52,8 +51,8 @@ func TestLogLevel(t *testing.T) {
 	}
 
 	for k, v := range examples {
-		t.Run(fmt.Sprintf("when LOGGER_LEVEL=%s", k), func(t *testing.T) {
-			os.Setenv("LOGGER_LEVEL", k)
+		t.Run("when LOGGER_LEVEL="+k, func(t *testing.T) {
+			t.Setenv("LOGGER_LEVEL", k)
```
leo-scalingo, line 55: I just learnt something here, thanks <3

logger/redacting_formatter.go
sc-zenokerr: It's a good question.
I tried to take an approach that would be configurable to each service.
If the Regexp is set to nil, the field will be redacted in it's entirety.

The reason the regexp is needed is for this issue: https://github.com/Scalingo/logs-service/issues/509 in which the path field (`path="/apps/66b24069fb0de6002981dd79/logs?timestamp=1727183062&token=[HERE]&stream=true"`) contains the token in the query string. In order to redact this, a regexp is required.

An alternative solution (perhaps better) is to avoid having the token in the query string in the first place, and move it to the request header instead, but that change is broader and has more impact.

What do you think is the better approach?

logger/redacting_formatter.go
sc-zenokerr: I did think about this. I hesitated, but you're right it's probably better so I'll add one.
Thanks for your input on this one!

logger/redacting_formatter.go
sc-david-voisin: Let's keep your approach and be vigilant at release !
The regexp you send from logs-service is compiled, this will help performance.

logger/redacting_formatter.go
sc-david-voisin: perfect :heart:

PR #997, opened 2024-11-20
package(mongo) [STORY-1030] Add option to define a customer method in order to pass custom query method (to look at paranoid-deleted items for instance

leo-scalingo:


mongo/pagination/pagination.go
```diff
@@ -24,11 +25,14 @@ type Meta struct {
...
 }
 
+type QueryFunc func(ctx context.Context, collectionName string, query bson.M, sortFields ...document.SortField) (*mgo.Query, document.Closer)
+
 type PaginateOpts struct {
 	PageNumber  int
 	AmountItems int
 	Query       bson.M
 	SortOrder   string
+	QueryFunc  QueryFunc
```
sc-zenokerr, line 35: Suggestion: add a comment to say what this property does. I need to read the code to find out that it replaces the default WhereQuery. It would also be good to know if it respects (or ignores) the "PageNumber", "AmountItems" and "SortOrder" properties.

mongo/pagination/pagination.go
```diff
@@ -83,7 +87,7 @@ func (s ServiceOpts) paramValidation(meta *Meta, collection string, opts *Pagina
 			append(badRequestErr.Errors[perPageErr], fmt.Sprintf("must be lower or equal to %d", s.MaxPerPage))
 	}
 
-	if badRequestErr.Errors != nil && len(badRequestErr.Errors) > 0 {
+	if len(badRequestErr.Errors) > 0 {
```
sc-zenokerr, line 90: praise: nice cleanup

mongo/pagination/pagination_test.go
```diff
@@ -338,6 +343,59 @@ func TestPaginationPaginate(t *testing.T) {
...
+					perPageNum:  10,
+				}
+			},
+			ExpectedResult: []dummyDocument{
+				{AppID: "0", VirtualStorageName: "vs_name_1"},
+				{AppID: "1", VirtualStorageName: "vs_name_1"},
+			},
+			ExpectedQuery: bson.M{"virtual_storage_name": "vs_name_1"},
+			ExpectedQueryFunc: document.WhereUnscopedQuery,
+		},
```
sc-zenokerr, line 398: Suggestion: Add extra tests to verify the behaviour of the query when used in combination with PageNumber, AmountItems and SortOrder properties.

sc-zenokerr (commented):
This looks good, but I think it misses a bit of documentation and some tests.
I think it is important to define the intended behaviour with regards to pagination and sort order and test for that behaviour. It "may" avoid confusion in the future. 
What do you think?

leo-scalingo:
What kind of tests would you add? In a way that the rest of the behavior is not changed, only the query getting the documents, that's why I only tested this. If you have one or two test cases to add I would definitely add them

Thanks for your review and I'll add the missing documentation

mongo/pagination/pagination.go
leo-scalingo: Ack

sc-zenokerr:
There are already a test cases for returning the second page:
* "It should return the second page"
* "It should return the second page in reverse order"

I would suggest adding equivalent tests when using the custom QueryFunc.

I think this would make it clear how we expect the library to behave with regards to pagination and sort order when using QueryFunc

What do you think?

mongo/pagination/pagination_test.go
leo-scalingo: Done

leo-scalingo:
> What do you think?

https://github.com/Scalingo/go-utils/pull/997/files#diff-e90d6bda755cb9479c9f732f823ac4aa98ebff14bc0bb7c1607d0d9bb5418cd6R115-R120

To me the change is minor enough to not having to redo the whole test matrix.
- Arguments are used the same way
- The library is not responsible of the WhereQuery anymore, if I would test more testcases, I would not test this change but `document.WhereUnscopedQuery`, a user of the library could build a custom `QueryFunc` which breaks the sorting because it takes the argument but does nothing with it, and then it wouldn't be the responsibility of the library. So I'm not sure it's pertinent to test all options.

PR #999, opened 2024-11-21
package(mongo) Do not break API to handle validation errors with internal errors

### LINKED ISSUE #998: package(mongo) Do not break Validate() API in next release, extend current API instead

leo-scalingo:
Fixes #998

### DIFF
--- a/mongo/document/document.go
+++ b/mongo/document/document.go
@@ -2,6 +2,7 @@ package document
 
 import (
 	"context"
+	stderrors "errors"
 	"fmt"
 	"time"
 
@@ -22,6 +23,7 @@ type document interface {
 	ensureID()
 	ensureCreatedAt()
 	setUpdatedAt(time.Time)
+	getUpdatedAt() time.Time
 	Validable
 }
 
@@ -46,8 +48,18 @@ type Closer interface {
 	Close()
 }
 
+var ErrValidateNoInternalErrorFunc = stderrors.New("no validation returning an internal error has been implemented")
+
 type Validable interface {
-	Validate(ctx context.Context) (*errors.ValidationErrors, error)
+	// Validate will be used if no ValidateWithInternalError is defined on a document
+	// It is not useful to have both defined on a document, only ValidationWithInternalError
+	// would be used in this case
+	Validate(ctx context.Context) *errors.ValidationErrors
+
+	// ValidateWithInternalError will be used in priority if defined on a document
+	// It will be called for all modifying operations (Create, Save, Update)
+	// If it returns an internal error, the validation error will be nil.
+	ValidateWithInternalError(ctx context.Context) (*errors.ValidationErrors, error)
 }
 
 var _ Validable = &Base{}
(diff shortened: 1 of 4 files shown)


EtienneM (changes requested):
I'm really not convinced by this API. Shouldn't we just revert #552? I mean, from the calling method, it's feasible to distinguish between a validation error and an internal error, isn't it?

leo-scalingo:
> I'm really not convinced by this API. Shouldn't we just revert #552? I mean, from the calling method, it's feasible to distinguish between a validation error and an internal error, isn't it?

I don't understand, currently we can' have a validation method which returns something else than a `*errors.ValidationErrors`, so it's just not possible to distinguish between a validation error and an internal error, because an internal error is not returnable.

My attempt here is to open to this possibility because indeed there are some simtuations in which a validation may fail because we can't access some data, or something internal is failing, with this PR it becomes possible without breaking the API. I thought it would be quite a elegant way to solve the initial problem which is indeed a problem.

leo-scalingo:
How would you handle this without breaking the API?

mongo/document/document.go
```diff
@@ -46,64 +48,85 @@ type Closer interface {
...
 }
 
+var ErrValidateNoInternalErrorFunc = stderrors.New("no validation returning an internal error has been implemented")
+
 type Validable interface {
-	Validate(ctx context.Context) (*errors.ValidationErrors, error)
+	// Validate will be used if no ValidateWithInternalError is defined on a document
+	// It is not useful to have both defined on a document, only ValidationWithInternalError
+	// would be used in this case
+	Validate(ctx context.Context) *errors.ValidationErrors
```
EtienneM, line 57: question: do we want to deprecate this method? In such case, we should document it IMO:

```suggestion
	// Deprecated: ...
	Validate(ctx context.Context) *errors.ValidationErrors
```

mongo/document/document.go
```diff
@@ -46,64 +48,85 @@ type Closer interface {
...
+		return c.UpdateId(doc.getID(), update)
+	})
+}
+
+func save(ctx context.Context, collectionName string, doc document, saveFunc func(context.Context, string, document) error) error {
+	validationErrors, err := doc.ValidateWithInternalError(ctx)
+	if err == ErrValidateNoInternalErrorFunc {
+		validationErrors = doc.Validate(ctx)
+	} else if err != nil {
+		return errors.Wrap(ctx, err, "fail to validate document")
```
EtienneM, line 119: nitpick(non-blocking):

```suggestion
		return errors.Wrap(ctx, err, "validate document")
```

mongo/document/document.go
```diff
@@ -46,64 +48,85 @@ type Closer interface {
...
-	if err != nil {
-		log.WithError(err).Error("Internal error while validating the document")
+func Save(ctx context.Context, collectionName string, doc document) error {
+	return save(ctx, collectionName, doc, func(ctx context.Context, collectionName string, doc document) error {
+		log := logger.Get(ctx)
+		c := mongo.Session(log).Clone().DB("").C(collectionName)
+		defer c.Database.Session.Close()
+		log.WithFields(logrus.Fields{
+			"collection": collectionName,
+			"doc_id":     doc.getID().Hex(),
```
EtienneM, line 89: suggestion: what do you think about adding these fields in the logger in the `save` function?

EtienneM (approved):
Sorry, I didn't understood correctly at the first read.

I'm already approving despite the suggestions and the failing CI. I let you choose whether you want to take into account the suggestions or not.

After releasing this, I think it would be nice to send a message on team-tech so that all Go developers are aware of this change.

leo-scalingo:
> I'm already approving despite the suggestions and the failing CI. I let you choose whether you want to take into account the suggestions or not.

I'll take t hem all into consideration :)

mongo/document/document.go
leo-scalingo: I don't think we want to deprecate, it's just 2 ways to define validation, I think that keeping a simple signature for the comon case works well.

mongo/document/document.go
leo-scalingo: Ack

PR #1061, opened 2025-02-04
feat(otel): add module for our OpenTelemetry SDK wrapper

### LINKED ISSUE #1044: [IBM] Initialize the OTEL Go package on go-utils

SCKevinO:
Fix #1044 


otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+}
+
+// GetGlobalMeter returns the global Meter
+func GetGlobalMeter() metric.Meter {
+	if globalWrapper == nil {
+		panic("Global OtelWrapper is not initialized")
+	}
+	return globalWrapper.globalMeter
+}
```
brandon-welsch, line 158: A global meter is not a great thing as we want to have multiple meter. Can we just handle `meter` in a separate PR.

otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+	counter, err := GetGlobalMeter().Float64Counter(
+		name,
+		metric.WithDescription(description),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create counter: %w", err)
+	}
+	counter.Add(ctx, value)
+	return nil
+}
```
brandon-welsch, line 171: Same for that, i would handle this in a separate PR as there's multiple issue. Step by step.

otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	switch cfg.ExporterType {
+	case "http":
+		return otlpmetrichttp.New(ctx)
+	case "grpc":
+		return otlpmetricgrpc.New(ctx)
+	default:
+		return nil, errors.New("invalid exporter type")
+	}
+}
```
brandon-welsch, line 150: Top 👍🏻

otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+	if cfg.MetricsExporter == nil {
+		exporter, err := newMetricsExporter(ctx, cfg)
+		if err != nil {
+			return nil, fmt.Errorf("failed to load exporter", err)
+		}
+		cfg.MetricsExporter = exporter
+	}
+	if cfg.MetricsReader == nil {
+		cfg.MetricsReader = sdkmetric.NewPeriodicReader(cfg.MetricsExporter, sdkmetric.WithInterval(cfg.CollectionInterval))
+	}
```
brandon-welsch, line 116: I'm not sure we want to override MetricReader. I'm not sure we want to use a ManualReader which add more things to do.
The only thing that need to be override is the collection interval if needed.

otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// Set up meter provider.
+	meterProvider, err := newMeterProvider(ctx, cfg)
+	if err != nil {
+		handleErr(err)
+		return
+	}
+	shutdownFuncs = append(shutdownFuncs, meterProvider.Shutdown)
+	otelsdk.SetMeterProvider(meterProvider)
```
brandon-welsch, line 92: I'm not sure this is needed, as we do our own SDK

otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+
+// InitSingleton initializes the MeterProvider as a singleton.
+// Subsequent calls to this function will return the same shutdown function and error.
+// The configuration is used only on the first call.
+func InitGlobalWrapper(ctx context.Context, cfg Config) (func(context.Context) error, error) {
+	globalOnce.Do(func() {
+		singletonShutdown, initErr = New(ctx, cfg)
+	})
+	return singletonShutdown, initErr
+}
```
brandon-welsch, line 63: For me this function is not needed. This can be done inside the `New` construct directly. It will remove complexity.

Only `New` should be exported here i think

otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+
+func New(ctx context.Context, cfg Config) (shutdown func(context.Context) error, err error) {
```
brandon-welsch, line 65: I would more see a `Shutdown` function here to handle `shutdownFuncs` in the service. Example:

```go
# main.go

err := otel.New(ctx)
if err != nil {
  // Handle error
}

defer otel.Shutdown()
```

This will give the ability to use `defer` in the main function, and handle properly the shutdown whenever the program is exited

otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+
+// OtelWrapper encapsulates OpenTelemetry MetricProvider and utilities
+type OtelWrapper struct {
+	meterProvider *sdkmetric.MeterProvider
+	globalMeter   metric.Meter
+	config        Config
+}
+
+var (
+	globalWrapper *OtelWrapper
```
brandon-welsch, line 43: I'm not fan of this name. We should need to find a better name.
I think the `Wrapper` term should be completely omitted in this Go package

otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// singletonShutdown holds the shutdown function once the meter provider is initialized.
+	singletonShutdown func(context.Context) error
+
+	// once ensures that initialization happens only once.
+	globalOnce sync.Once
+
+	// initErr captures any error encountered during initialization.
+	initErr error
```
brandon-welsch, line 52: This is not needed i think, not globally. error should be handled in `New` and `Shutdown` functions instead.

otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+// Config defines the Otel Wrapper configuration for metrics
+type Config struct {
+	ServiceName               string
+	Debug                     bool
+	ExporterType              string
+	ExporterEndpoint          string
+	ExporterTimeout           string
+	ExporterCertificate       string
+	ExporterClientCertificate string
+	ExporterClientKey         string
```
brandon-welsch, line 29: I think that there's so much thing in the same Config struct and that we need to make something like:
```go
type ExporterConfig {
  	Type              string
	Endpoint          string
	Timeout           string
	Certificate       string
	ClientCertificate string
	ClientKey         string
}

type ReaderConfig [
  Interval int
}

type Config struct {
  ServiceName string
  Debug bool
  Exporter ExporterConfig
  Reader ReaderConfig
}
```

This should also use environment tag to be correctly initialized from environment variables, but let's make that in another PR

otel/otel.go
```diff
@@ -0,0 +1,171 @@
...
+	CollectionInterval        time.Duration
+	MetricsReader             sdkmetric.Reader
+	MetricsExporter           sdkmetric.Exporter
```
brandon-welsch, line 32: I also think that these information should not be keeped in Config struct. But in the struct that will be keeped globally through the  `singleton`

brandon-welsch (changes requested):

otel/otel.go
SCKevinO: I think it can be worth it to keep it. If we want to use some of the automatic instrumentations available in the official Go SDK, they might be relying on this.

brandon-welsch: This can be added later, currently it's not in the scope.

And plus of that actually from what i see instrumentation is registered like that:
```go
runtime.Start(runtime.WithMeterProvider(meterProvider))
```

So it's kind of useless for now

brandon-welsch: But let's go, keep it like that

brandon-welsch (approved):

PR #1067, opened 2025-02-07
test(otel): Add unit test for otel package

### LINKED ISSUE #1065: [IBM] OTEL-SDK: Add unit tests on the package

brandon-welsch:
Fix #1065


otel/otel_test.go
```diff
@@ -0,0 +1,98 @@
...
+		},
+		{
+			name:   "re-initialization to check singleton usage",
+			reinit: true,
+			env: map[string]string{
+				"OTEL_SERVICE_NAME": "test",
+			},
+		},
+		{
+			name: "initialization with debug enabled",
```
SCKevinO, line 37: this is not tested below, is that normal?

otel/otel_test.go
```diff
@@ -0,0 +1,98 @@
...
+			env: map[string]string{
+				"OTEL_SERVICE_NAME": "test",
+				"OTEL_DEBUG":        "true",
+			},
+		},
+		{
+			name: "initialization with grpc exporter",
```
SCKevinO, line 44: idem

otel/otel_test.go
```diff
@@ -0,0 +1,98 @@
...
+			env: map[string]string{
+				"OTEL_SERVICE_NAME":  "test",
+				"OTEL_EXPORTER_TYPE": "grpc",
+			},
+		},
+		{
+			name: "initialization with collection interval at 30s",
```
SCKevinO, line 51: idem

otel/otel_test.go
brandon-welsch: It's inside the OTEL package stored as private. And there's no mocks in fact for OTEL `metrics` package

brandon-welsch: I could try to see if can can generate mock for the OTEL SDK

SCKevinO: ok, in that case I don't think we need to define the test cases at all here, it's a bit misleading as someone looking at the tests could think this is tested and working as expected when changes things in the code

SCKevinO (approved):

PR #1073, opened 2025-02-10
feat(otel): create a meter

### LINKED ISSUE #1063: [IBM] OTEL-SDK-4: Support creating a Meter easily
This should implement the following spec: https://www.notion.so/scalingooriginal/Support-creating-a-Meter-easily-184f536d89bf80dc8e55e7190a5f1f25

SCKevinO:
Fix #1063 

### DIFF
--- a/otel/metrics.go
+++ b/otel/metrics.go
@@ -1,0 +1,25 @@
+package otel
+
+import (
+	sdkmetric "go.opentelemetry.io/otel/metric"
+)
+
+type Meter interface {
+	GetOriginalMeter() sdkmetric.Meter
+}
+
+type meter struct {
+	name     string
+	sdkMeter sdkmetric.Meter
+}
+
+func GetMeter(name string) Meter {
+	return &meter{
+		name:     name,
+		sdkMeter: globalProviders.meterProvider.Meter(name),
+	}
+}
+
+func (m *meter) GetOriginalMeter() sdkmetric.Meter {
+	return m.sdkMeter
+}
--- a/otel/metrics.go
+++ b/otel/metrics.go
@@ -4,9 +4,7 @@ import (
 	sdkmetric "go.opentelemetry.io/otel/metric"
 )
 
-type Meter interface {
-	GetOriginalMeter() sdkmetric.Meter
-}
+type Meter interface{}
 
 type meter struct {
 	name     string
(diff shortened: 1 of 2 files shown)


otel/metrics.go
```diff
@@ -0,0 +1,25 @@
+package otel
+
+import (
+	sdkmetric "go.opentelemetry.io/otel/metric"
+)
+
+type Meter interface {
+	GetOriginalMeter() sdkmetric.Meter
```
brandon-welsch, line 8: We don't need this `GetOriginalMeter()`. The interface is needed for the next PR to list all methods to create instruments
```suggestion
```

otel/metrics.go
```diff
@@ -0,0 +1,25 @@
...
+func GetMeter(name string) Meter {
+	return &meter{
+		name:     name,
+		sdkMeter: globalProviders.meterProvider.Meter(name),
+	}
+}
+
+func (m *meter) GetOriginalMeter() sdkmetric.Meter {
+	return m.sdkMeter
+}
```
brandon-welsch, line 25: We don't need this. As instruments methods that will be implemented in the next PR, will use the meter internally through the struct. And i'm not sure we need to get this meter outside for now

brandon-welsch (changes requested):

otel/metrics.go
SCKevinO: It's just a way to escape the abstraction if needed. That way, if for some reason, the methods we will implement are not enough, we can still get a meter and use the official sdk.
It should not be used in most cases but I think it's better to have it, just in case

brandon-welsch: Yes but we could add it later if we need it. Actually keeping it minimal is better

SCKevinO: I'd say this PR is already pretty minimal as it is. Why remove this one-line function to add it back later? The otel sdk is quite complex and I can't pretend that we will handle every use case in this project, so an escape hatch can be useful

brandon-welsch: I think it would prefer to don't mix our otel package with the one of opentelemetry now. If it's not used properly later, it could add some complexity. 

We need a minimal package that works the way we want it to. The marginal cases we'll see later. The goal for now is to have a working package.

It should be simple and easy to use, so for now i don't want to keep this as we don't have a case where use it for now. And honestly the name of the methods doesn't satisfy me also.

SCKevinO: Fine I'll remove it. I don't agree with this but I don't want to spend all day arguing about it...

brandon-welsch: For now, just make this package work as we want. We will see that later

brandon-welsch (approved):

PR #1077, opened 2025-02-11
feat(s3) Add method GetWithRetries to stream an object to a writer and retry download if it is interrupted or if the all data couldn't be read

### LINKED ISSUE #1076: module(storage) S3 Get() shuld have a way to retry connection to backend in case it is cut for any reason
https://github.com/aws/aws-sdk-go/issues/4510
https://alexwlchan.net/2019/streaming-large-s3-objects/

leo-scalingo:
Fixes #1076

### DIFF
--- a/storage/CHANGELOG.md
+++ b/storage/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## To be Released
 
+## v1.4.0
+
+* feat(s3): Add method `GetWithRetries` taking a writer as argument and which
+  will handle gracefully if a `GetObject` sent to the object storage is failing
+  and has to be retried.
+
 ## v1.3.3
 
 * build(deps): bump github.com/aws/aws-sdk-go-v2/feature/s3/manager from 1.11.37 to 1.11.46
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -62,29 +62,29 @@ type S3 struct {
 	partSize          int64
 }
 
-type s3Opt func(s3 *S3)
+type S3Opt func(s3 *S3)
 
 // WithRetryPolicy is an option to constructor NewS3 to add a Retry Policy
 // impacting GET operations
-func WithRetryPolicy(policy RetryPolicy) s3Opt {
-	return s3Opt(func(s3 *S3) {
+func WithRetryPolicy(policy RetryPolicy) S3Opt {
+	return S3Opt(func(s3 *S3) {
 		s3.retryPolicy = policy
 	})
 }
 
-func WithPartSize(size int64) s3Opt {
-	return s3Opt(func(s3 *S3) {
+func WithPartSize(size int64) S3Opt {
+	return S3Opt(func(s3 *S3) {
 		s3.partSize = size
 	})
 }
 
-func WithUploadConcurrency(concurrency int) s3Opt {
-	return s3Opt(func(s3 *S3) {
+func WithUploadConcurrency(concurrency int) S3Opt {
+	return S3Opt(func(s3 *S3) {
 		s3.uploadConcurrency = concurrency
 	})
 }
 
-func NewS3(cfg S3Config, opts ...s3Opt) *S3 {
+func NewS3(cfg S3Config, opts ...S3Opt) *S3 {
 	s3config := s3Config(cfg)
 	s3client := s3.NewFromConfig(s3config)
 	s3 := &S3{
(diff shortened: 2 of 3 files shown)


storage/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## To be Released
 
+## v1.4.0
```
EtienneM, line 5: The version number shouldn't be included yet. It will be included as part of the release process (https://github.com/Scalingo/go-utils?tab=readme-ov-file#release-a-new-version-of-a-module).

Why not having the release process and the feature addition in a single PR. Then you need to update the module README.

storage/s3.go
```diff
@@ -126,13 +126,72 @@ func (s *S3) Get(ctx context.Context, path string) (io.ReadCloser, error) {
...
+	if err != nil {
+		return -1, errors.Wrapf(err, "get size of S3 object %v", path)
+	}
+
+	var (
+		offset  int64
+		attempt = 1
+	)
+
+	for {
```
EtienneM, line 157: question: I'm not keen on this apparently infinite loop. Would it be possible to have a failsafe with a max number of attempt or something like that? Do we want to use the retryer (https://github.com/Scalingo/go-utils/tree/master/retry) for this use case?

storage/s3.go
```diff
@@ -126,13 +126,72 @@ func (s *S3) Get(ctx context.Context, path string) (io.ReadCloser, error) {
...
 	}
 	return out.Body, nil
 }
 
+// GetWithRetries function downloads the object from S3 and writes it to the
+// writer. It returns the number of bytes written to the writer.
+//
+// It implements a retry mechanism to handle if connection is closed before
+// the end of the download.
+func (s *S3) GetWithRetries(ctx context.Context, path string, writer io.Writer) (int64, error) {
```
EtienneM, line 142: suggestion(non-blocking): I think that in Go, method writing things to a `io.Writer` are named with the word `Write` somewhere
```suggestion
func (s *S3) WriteWithRetries(ctx context.Context, path string, writer io.Writer) (int64, error) {
```

storage/s3.go
```diff
@@ -126,13 +126,72 @@ func (s *S3) Get(ctx context.Context, path string) (io.ReadCloser, error) {
...
+
+	var (
+		offset  int64
+		attempt = 1
+	)
+
+	for {
+		// Log if we are resuming a download
+		if offset != 0 {
+			log.Infof("Downloading from offset %d...", offset)
```
EtienneM, line 160: suggestion(non-blocking): add the offset as a logger field
```suggestion
			log.WithField("offset", offset).Infof("Downloading from offset %d...", offset)
```

storage/s3.go
```diff
@@ -126,13 +126,72 @@ func (s *S3) Get(ctx context.Context, path string) (io.ReadCloser, error) {
...
+		})
+		if err != nil {
+			if attempt >= s.retryPolicy.Attempts {
+				return -1, errors.Wrap(err, "get object after retries")
+			}
+			time.Sleep(s.retryPolicy.WaitDuration)
+			log.Infof("Get Object request error: %v, retrying", err)
+			attempt++
+			continue
+		}
```
EtienneM, line 178: question: in this issue (https://github.com/aws/aws-sdk-go/issues/4510), it is mentioned that the error is EOF when failing to download a big file. Should we only retry in case of EOF and fail in other cases?

storage/s3.go
```diff
@@ -126,13 +126,72 @@ func (s *S3) Get(ctx context.Context, path string) (io.ReadCloser, error) {
...
+			Bucket: &s.cfg.Bucket,
+			Key:    &path,
+			Range:  aws.String(rangeHeader),
+		})
+		if err != nil {
+			if attempt >= s.retryPolicy.Attempts {
+				return -1, errors.Wrap(err, "get object after retries")
+			}
+			time.Sleep(s.retryPolicy.WaitDuration)
+			log.Infof("Get Object request error: %v, retrying", err)
```
EtienneM, line 175: suggestion(non-blocking): errors should be logged as logger fields
```suggestion
			log.WithError(err).Info("Get Object request error, retrying")
```

storage/s3.go
```diff
@@ -126,13 +126,72 @@ func (s *S3) Get(ctx context.Context, path string) (io.ReadCloser, error) {
...
+			continue
+		}
+		defer resp.Body.Close()
+
+		// Stream object data to response, tracking bytes written
+		bytesWritten, err := io.Copy(writer, resp.Body)
+		if err != nil || bytesWritten < offsetSize {
+			// Handle partial writes and retry
+			offset += bytesWritten
+			log.Infof("Streaming error: %v. Retrying from offset %d...", err, offset)
```
EtienneM, line 186:
```suggestion
			log.WithError(err).WithField("offset", offset).Infof("Streaming error, retrying from offset %d...", offset)
```

storage/CHANGELOG.md
leo-scalingo: You're right!

storage/s3.go
leo-scalingo: Actually it's because I wanted to keep the scope of the lib, we're doing a s3Client.Get(), to get a file from S3, doing s3lcient.Write would be misleading on the API usage, we're looking to download a file from remote

storage/s3.go
leo-scalingo: The retryer doesn't work well here, because there are 2 kinds of retries which are intertwined. HTTP request to the backend (which are counted as hard failure) and data streaming from the Body.

In the latter, we'll want to retry until we got the whole file. Since we don't control network, when downloading 500GB, it can be stopped every 10GB or 50GB or 100GB and we'll want to keep on.

That's the reason I've done it that way.

storage/s3.go
leo-scalingo: Ack

storage/s3.go
leo-scalingo: It's not EOF when doing the `GetObject`, but when reading the body. And actually when reading the body, we can't have a lot of different errors, EOF/Unexpected EOF/Short Read (when what is read doesn't match what is expected to be read), and in this case I think it's ok to match all the errors, once the initial HTTP request has returned a success.

storage/s3.go
leo-scalingo: Ack

storage/s3.go
```diff
@@ -126,13 +126,80 @@ func (s *S3) Get(ctx context.Context, path string) (io.ReadCloser, error) {
...
 		return nil, errors.Wrapf(err, "fail to get S3 object %v", path)
 	}
 	return out.Body, nil
 }
 
+// GetWithRetries function downloads the object from S3 and writes it to the
+// writer. It returns the number of bytes written to the writer.
+//
+// It implements a retry mechanism to handle if connection is closed before
+// the end of the download.
```
EtienneM, line 141: From what I understand from the updated PR title, this comment is no longer accurate
```suggestion
// It implements a retry mechanism to handle if connection is closed before
// the end of the download or if all data couldn't be read.
```

EtienneM (approved):
That's great! I disabled auto-merge because I have one last suggestion. But I'm already approving the PR. Feel free to take it into account or not.

leo-scalingo:
> That's great! I disabled auto-merge because I have one last suggestion. But I'm already approving the PR. Feel free to take it into account or not.

Thanks a lot

PR #1086, opened 2025-02-24
feature(s3) Use the retrier options to configure the AWS SDK

john-scalingo:
Related to https://github.com/Scalingo/project-items/issues/785

The issue with the current retrier is that it only works on simple retryable operations (like Size, Info, etc.). However on complex operations (like a Multipart upload) we implemented no retries and did let the default AWS SDK options (3 retries).
This PR aims to align the SDK retry options with the ones configured on our custom retrier.

### DIFF
--- a/storage/CHANGELOG.md
+++ b/storage/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## To be Released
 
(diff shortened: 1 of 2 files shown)


leo-scalingo (approved):
Thanks for the PR, LGTM

PR #1092, opened 2025-02-28
fix(otel): Use OTEL_METRIC_EXPORT_INTERVAL instead of OTEL_COLLECTION_INTERVAL env var

### LINKED ISSUE #1091: [otel] Remove OTEL_COLLECTION_INTERVAL env var in favor of OTEL_METRIC_EXPORT_INTERVAL
Reference: https://github.com/open-telemetry/opentelemetry-go/blob/a9cbc3d8dec7be22c7d3691ca1755f25c1702a1d/sdk/metric/env.go

brandon-welsch:
Fix #1091 

### DIFF
--- a/otel/otel.go
+++ b/otel/otel.go
@@ -21,7 +21,7 @@ type Config struct {
 	Debug                bool          `default:"false"`
 	ExporterType         string        `default:"http" split_words:"true"`
 	ExporterOtlpEndpoint string        `default:"" split_words:"true"`
-	CollectionInterval   time.Duration `default:"10s" split_words:"true"`
+	MetricExportInterval time.Duration `default:"10s" split_words:"true"`
 }
 
 func Init(ctx context.Context) (func(context.Context) error, error) {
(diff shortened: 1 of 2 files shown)


otel/otel.go
```diff
@@ -17,11 +16,10 @@ import (
 )
 
 type Config struct {
-	ServiceName          string        `required:"true" split_words:"true"`
-	Debug                bool          `default:"false"`
-	ExporterType         string        `default:"http" split_words:"true"`
-	ExporterOtlpEndpoint string        `default:"" split_words:"true"`
-	CollectionInterval   time.Duration `default:"10s" split_words:"true"`
```
SCKevinO, line 24: I would keep this (but change the name to match the otel name). This way, we can control the default value instead of relying on the otel one.

brandon-welsch: If you want. But honestly i prefer that we control the default value through the env through Terraform or Chef when we deploy apps/services.

I don't want to have to update the package, make a release of this package then update all apps/service in order to take the default value.

For me it make sense to have this default value in Terraform modules or Chef default attributes

brandon-welsch: WDYT of that way instead of handling the default value here

SCKevinO: I agree that it's better to define the value with an env var. And here the default is only applied in case the env var is missing. So we can still use the method you describe with terraform/chef but in case the env var is missing, we have a default value that we control.

brandon-welsch: Okay, let's go with that

otel/otel.go
```diff
@@ -40,7 +40,10 @@ func Init(ctx context.Context) (func(context.Context) error, error) {
 		return nil, errors.Wrap(ctx, err, "load exporter")
 	}
 
-	metricsReader := sdkmetric.NewPeriodicReader(metricsExporter, sdkmetric.WithInterval(cfg.CollectionInterval))
```
SCKevinO, line 43: you also need to keep this

SCKevinO: sorry, didn't see the full diff, my bad

PR #1123, opened 2025-03-17
fix(graceful): VULN Defer mutex unlock + lint

sc-zenokerr:
This PR fixes issue #1122 identified by semgrep, where a mutex lock might not be released

- [x] Add a changelog entry in `CHANGELOG.md`

PR #1124, opened 2025-03-17
fix(graceful/logger): revert erroneous changelog update

sc-zenokerr:
This PR fixes an error merged in the PR #1123 where the incorrect changelog was updated.


leo-scalingo (approved):
LGTM

PR #1153, opened 2025-05-06
feat(retry) when WithMaxDuration is used, it ignores the default MaxAttempts

curzolapierre:

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -68,33 +68,53 @@ func WithMaxAttempts(maxAttempts int) RetryerOptsFunc {
 	}
 }
 
+// WithMaxDuration sets the timeout which will cancel the retry loop if it has
+// expired. It does not affect the timeout of the method given in the Do method.
+// When MaxDuration is sets but not MaxAttempts, the number of attempts is
+// math.MaxInt32.
 func WithMaxDuration(duration time.Duration) RetryerOptsFunc {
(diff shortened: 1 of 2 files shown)


leo-scalingo (commented):
So if I do:

```
r = NewRetrier(WithoutMaxAttempt())
```

It will be set to 5, not really logical unfortunately

curzolapierre:
> It will be set to 5, not really logical unfortunately

Oh indeed, this case is not covered in tests.

So we face to the issue of default value here.
Or what about returning an error in this particular case

What would be the use case of retrier without `MaxAttempts`?
Does this mean we are expecting an infinite retry? Or we do necessarily want a limit, in which case we need to specify it

leo-scalingo:
Actually I think it's better to 
- Have good defaults
- Be explicit in the API, which means, no side effect

The condition you're adding is a kind of side effect which may lead to situations which are difficult to understood without reading the doc/code of the library.

So to me the question is: 
- Is the current default un-practical and we want to change it?
- What would be the ideal explicit API for you?

leo-scalingo (commented):
See above comment?

EtienneM:
@curzolapierre I feel like you forgot about this one. Can you revive this PR?

EtienneM:
@curzolapierre is it still under your radar?

PR #1164, opened 2025-06-03
[STORY-2085] Update Go version to 1.24.3

briceamen:
This PR Update Go version of all modules to 1.24.3 in order to fix golangci-lint.

### DIFF
--- a/concurrency/go.mod
+++ b/concurrency/go.mod
@@ -1,6 +1,6 @@
 module github.com/Scalingo/go-utils/concurrency
 
-go 1.23.3
+go 1.24.3
 
 require github.com/Scalingo/go-utils/logger v1.5.0
 
--- a/cronsetup/go.mod
+++ b/cronsetup/go.mod
@@ -1,8 +1,6 @@
 module github.com/Scalingo/go-utils/cronsetup
 
-go 1.23.0
-
-toolchain go1.24.1
+go 1.24.3
 
 require (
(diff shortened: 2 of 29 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# CODEOWNERS for Scalingo/go-utils
+# This file defines who is responsible for code reviews in this repository
+
+# Global code owners - these users will be requested for review on all PRs
+* @EtienneM
```
curzolapierre, line 5: We don't want the whole PRs to be assigned to Étienne but only Dependabot's ones.

It should be something like the following suggestion:
```suggestion
# Go
go.sum @EtienneM

# Github Actions
.github/workflows/* @EtienneM
```

But take a look to the Github CODEOWNERS documentation, maybe `go.sum @<user>` is not enough and should be instead something more explicit:
```
concurrency/go.sum
cronsetup/go.sum
...
```
Or even better string match?
```
*/go.sum
```

curzolapierre (changes requested):

.github/CODEOWNERS
briceamen: Ok I get it ! Thanks !

briceamen: Based on the doc : https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

We should use :
```
**/go.sum
```
To match at any depth.

Issue #1165, opened 2025-06-05
[IBM] Add an option to completely disable the otel sdk


SCKevinO:
In some cases, for example integration tests, we need to disable the otel sdk.
The OpenTelemetry spec contains an environment variable OTEL_SDK_DISABLED that do just that but the go implementation chose to not implement it (see https://github.com/open-telemetry/opentelemetry-go/issues/3559).
It would be nice to add this option in our own sdk.

PR #1166, opened 2025-06-05
fix(otel): add OTEL_SDK_DISABLED env var

SCKevinO:
Fix https://github.com/Scalingo/go-utils/issues/1165

### DIFF
--- a/otel/otel.go
+++ b/otel/otel.go
@@ -19,6 +19,7 @@ import (
 type Config struct {
 	ServiceName          string        `required:"true" split_words:"true"`
 	Debug                bool          `default:"false"`
+	SdkDisabled          bool          `default:"false" split_words:"true"`
 	DebugPrettyPrint     bool          `default:"true" split_words:"true"`
 	ExporterType         string        `default:"http" split_words:"true"`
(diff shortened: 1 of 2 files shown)


otel/otel.go
```diff
@@ -33,6 +34,12 @@ func Init(ctx context.Context) (func(context.Context) error, error) {
 		return nil, errors.Wrap(ctx, err, "load configuration")
 	}
 
+	if cfg.SdkDisabled {
```
brandon-welsch, line 37: Did you tried, that it is working on swan-agent? I mean that if i disable it on the whole app, then run the app, calls to send metrics is not impacted?

SCKevinO: I tested it with my test app on staging (https://admin.osc-st-fr1.st-sc.fr/apps/65034575884fef000f77180c).
When we add OTEL_SDK_DISABLED=true in the env, we skip the initialization and all subsequent calls to otel are no-op.
On my test app, it worked great, no logs appeared and no data showed up in grafana

brandon-welsch: Nice perfect 👌🏻LGTM so

PR #1174, opened 2025-06-06
[STORY-2085] Missing any depth checking for go.sum

briceamen:
The reason for changing from */go.sum to **/go.sum is that we have go.sum files at different depth levels in our repository:

  - Most go.sum files are at depth 2 (e.g., ./logger/go.sum, ./crypto/go.sum)
  - But we also have go.sum files at depth 5 in the otel examples:
    - ./otel/docs/examples/int64-async-gauge/go.sum
    - ./otel/docs/examples/int64-sync-counter/go.sum

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,7 @@
 # This file defines who is responsible for code reviews in this repository
 
 # Go dependency files - assign to EtienneM for Dependabot PRs
-*/go.sum @EtienneM
+**/go.sum @EtienneM
 
 # GitHub Actions workflows
 .github/workflows/* @EtienneM


curzolapierre (commented):
Could you please explain this change?
What is the use case we would need to check deep go.sum?

curzolapierre (commented):
In your example, `./logger/go.sum` is matching `*/go.sum`.

For `./otel/docs/examples/int64` we are in case of an example. Do you think dependencies should be handled by Dependabot?
If you don't have the answer, ping SRE as they have maybe miss this one.

In addition, I noticed that this package `otel` is not covered by Dependabot (cf: `.github/dependabot.yml`).

briceamen:
> In your example, `./logger/go.sum` is matching `*/go.sum`.
> 
> For `./otel/docs/examples/int64` we are in case of an example. Do you think dependencies should be handled by Dependabot? If you don't have the answer, ping SRE as they have maybe miss this one.
> 
> In addition, I noticed that this package `otel` is not covered by Dependabot (cf: `.github/dependabot.yml`).

Ok I am closing this PR, this example doesn't need to be covered by Dependabot

briceamen:
In depth not required

PR #1229, opened 2025-07-11
[STORY-2247] Add ability to log a failed attempt instead of failing completely silently

leo-scalingo:


retry/retry_test.go
```diff
@@ -181,4 +185,23 @@ func TestRetrier(t *testing.T) {
...
+			WithWaitDuration(10*time.Millisecond),
+			WithLoggingOnAttemptError(logrus.ErrorLevel),
+		)
+
+		retrier.Do(ctx, func(ctx context.Context) error {
+			return errors.New("TestError")
+		})
+
+		assert.Len(t, hook.Entries, 2)
+		assert.Contains(t, hook.Entries[0].Message, "attempt failed")
```
matthieu526-scalingo, line 204: There is an error in the order of parameters, it should be `expectec, actual`.
```suggestion
		assert.Contains(t, "attempt failed", hook.Entries[0].Message)
```

retry/retry.go
```diff
@@ -86,6 +90,16 @@ func WithErrorCallback(c ErrorCallback) RetryerOptsFunc {
 	}
 }
 
+func WithLoggingOnAttemptError(severity logrus.Level) RetryerOptsFunc {
```
matthieu526-scalingo, line 93: Should the `severity` be sent as parameter, as should it be always `info`?
If it can be sent, I would add a comment stating that if the error level is set to `error` Rollbar will be "poluted" with intermediate steps.

matthieu526-scalingo (approved):

PR #1254, opened 2025-07-16
build(dependabot): ignore `github.com/aws/aws-sdk-go-v2` without version specified

EtienneM:
We try to configure Dependabot to ignore this Dependency. For some reason, the previous configuration (#1236) didn't work (#1251). Let's try to configure this without specifying the version to see if it works...


EtienneM:
Sorry Pierre, for some tests I re-created the PR...

PR #1277, opened 2025-08-05
docs(otel): Update examples to show a better naming for attribute names

### LINKED ISSUE #1269: [IBM] Update examples in otel package
Example: https://github.com/Scalingo/go-utils/blob/master/otel/docs/examples/int64-async-gauge/metric/database.go#L12

Don't show a bad example that define `env` resource attribute. This kind of resource attribute shouldn't be defined at this level on metric. Change it to show a better example.

brandon-welsch:
Fix #1269


stopostit (approved):
LGTM

PR #1280, opened 2025-08-18
fix(otel): Return early if SDK is disabled

### LINKED ISSUE #1279: [otel] If SDK is disabled it should exit earlier, before configuration is loaded

brandon-welsch:
Fix #1279 

### DIFF
--- a/otel/otel.go
+++ b/otel/otel.go
@@ -19,6 +19,7 @@ import (
 	"google.golang.org/grpc/credentials"
 
 	"github.com/Scalingo/go-utils/errors/v2"
+	"github.com/Scalingo/go-utils/logger"
 )
 
 type Config struct {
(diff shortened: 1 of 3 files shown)


briceamen (approved):
LGTM

otel/otel.go
```diff
@@ -34,19 +34,21 @@ type Config struct {
 }
 
 func Init(ctx context.Context) (func(context.Context) error, error) {
+	// If SDK is disabled through env, exit earlier without any error
+	isSDKDisabled := os.Getenv("OTEL_SDK_DISABLED")
+	if isSDKDisabled == "true" {
+		return func(ctx context.Context) error {
+			return nil
```
curzolapierre, line 41: **Suggestion**: IMO a log message would be useful here

curzolapierre (commented):
**Question**: would we prefer to explicit disable the initialization or more if the required env variables are not set then the service is not init

PR #1293, opened 2025-08-27
fix(nsqconsumer/error) in case of nsqconsumer.Error we unwrap two times

curzolapierre:
So we may miss a errCtx containing log fields.

~~This PR depends on https://github.com/Scalingo/go-utils/pull/1294 to works.~~ _Edit: Done_

Related to https://github.com/Scalingo/appsdeck-database/pull/3629

Fixes https://github.com/Scalingo/appsdeck-database/issues/3620

### DIFF
--- a/nsqconsumer/consumer.go
+++ b/nsqconsumer/consumer.go
@@ -14,7 +14,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/stvp/rollbar"
 
-	"github.com/Scalingo/go-utils/errors/v2"
+	"github.com/Scalingo/go-utils/errors/v3"
 	"github.com/Scalingo/go-utils/logger"
 	"github.com/Scalingo/go-utils/nsqproducer"
 )
--- a/nsqconsumer/go.mod
+++ b/nsqconsumer/go.mod
@@ -3,7 +3,7 @@ module github.com/Scalingo/go-utils/nsqconsumer
 go 1.24
 
 require (
-	github.com/Scalingo/go-utils/errors/v2 v2.5.1
+	github.com/Scalingo/go-utils/errors/v3 v3.1.0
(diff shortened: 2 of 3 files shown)


leo-scalingo (commented):
What's the source issue, where has this been seen? Do you have problematic logs somewhere?

Or I would need a test case or something. I don't know how to validate this change without any of that :-/

curzolapierre:
Yes sorry I don't linked the correct link from PR's description. You can find the related issue mentioned by SRE.

nsqconsumer/consumer.go
```diff
@@ -311,7 +311,6 @@ func (c *nsqConsumer) nsqHandler(message *nsq.Message) (err error) {
 				errLogger = logger.Get(handlerErr.Ctx())
 			case Error:
 				noRetry = handlerErr.noRetry
-				unwrapErr = handlerErr.error
```
leo-scalingo, line 314: No sure this patch is good,  `unwrapErr` is not unwrapped initially (in the first run of loop` 

`errors.UnwrapError(<nsqconsumer.Error>)` does return  `nil` am I wrong on this? Does it implement an interface like required here https://github.com/Scalingo/go-utils/blob/master/errors/cause.go#L47 ?

curzolapierre: `unwrapErr` is not unwrapped initially like was done before.

However, you're right it returns nil because `errors.UnwrapError` does not implement interface.

I'm suggesting this option: https://github.com/Scalingo/go-utils/compare/0c790a03b3d2d685c8dd5d20c9375a2280b421cd..ea7cb72593fdc1b6068e46a8e3d4238073db7ba5

curzolapierre: Yes nice idea.
Here the PR: https://github.com/Scalingo/go-utils/pull/1294

PR #1325, opened 2025-09-02
fix(otel): Switch resource to schemaless to avoid schema/semconv mismatch issues

### LINKED ISSUE #1324: [otel] Switch to schemaless, instead of defining a schema version
Currently we don't use strongly what is in the semconv. We use some attributes from it, but these ones are marked as `stable` and will not change as quite common. Thus we don't need to specify a specific version of semconv on the OpenTelemetry pipelines. 

This would let Dependabot upgrade the version of the Go SDK of OpenTelemetry without generating issues of compatibility.

brandon-welsch:
Fix #1324 

This will avoid issue when Dependabot upgrade OTEL SDK like: https://github.com/Scalingo/go-utils/pull/1319

### DIFF
--- a/otel/otel.go
+++ b/otel/otel.go
@@ -96,8 +96,10 @@ func Init(ctx context.Context) func() error {
 	res, err := resource.Merge(
 		resource.Default(),
 		resource.NewSchemaless(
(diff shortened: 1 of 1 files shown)


stopostit (approved):
LGTM

PR #1329, opened 2025-09-24
[STORY-2516] Add ability to define global attributes from caller with an option, including an helper to define Service Version

leo-scalingo:

### DIFF
--- a/errors/cause_test.go
+++ b/errors/cause_test.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	errgo "gopkg.in/errgo.v1"
 )
 
--- a/errors/cause_test.go
+++ b/errors/cause_test.go
@@ -178,7 +178,8 @@ func Test_UnwrapError(t *testing.T) {
 		}
 
 		assert.Equal(t, "test=biniou", UnwrapError(err).Error())
-		assert.IsType(t, &ValidationErrors{}, UnwrapError(err))
+		verr := &ValidationErrors{}
+		require.ErrorAs(t, UnwrapError(err), &verr)
 		assert.ErrorContains(t, err, "custom error test=biniou")
 	})
 
@@ -207,7 +208,8 @@ func Test_UnwrapError(t *testing.T) {
 			lastErr = unwrappedErr
 		}
 		assert.Equal(t, "test=biniou", lastErr.Error())
-		assert.IsType(t, &ValidationErrors{}, lastErr)
+		verr := &ValidationErrors{}
+		require.ErrorAs(t, lastErr, &verr)
 	})
 
 	t.Run("given a nil error", func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


otel/otel.go
```diff
@@ -95,12 +126,9 @@ func Init(ctx context.Context) func() error {
...
 	res, err := resource.Merge(
 		resource.Default(),
-		resource.NewSchemaless(
-			// https://opentelemetry.io/docs/specs/semconv/resource/#service
-			semconv.ServiceName(cfg.ServiceName),
-			semconv.ServiceInstanceID(serviceInstanceID),
-			// https://opentelemetry.io/docs/specs/semconv/resource/host/
-			semconv.HostName(hostName),
+		resource.NewWithAttributes(
+			semconv.SchemaURL,
```
brandon-welsch, line 130: Can you replace `resource.NewWithAttributes` by `resource.NewSchemaless` since it caused issue when Dependabot upgraded OTEL on services, because the SDK was not up-to-date.

brandon-welsch: Since `service.version` is stable (https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/#service-attributes), it shouldn't be an issue later

brandon-welsch (commented):
Nice addition. Thanks for the contribution.

However i think actually it would be better to use SchemaLess, as we had issues with dependabot updates on some project recently due to the mismatch of OTEL sdk version between this package and services that use it.

Can you also update examples in `otel/docs/examples`, in order to give an example on how to use `WithServiceVersionAttribute`?

errors/cause_test.go
```diff
@@ -207,7 +208,8 @@ func Test_UnwrapError(t *testing.T) {
 			lastErr = unwrappedErr
 		}
 		assert.Equal(t, "test=biniou", lastErr.Error())
-		assert.IsType(t, &ValidationErrors{}, lastErr)
+		verr := &ValidationErrors{}
+		require.ErrorAs(t, lastErr, &verr)
```
brandon-welsch, line 212: Why this is here? Shouldn't be in another PR?

otel/otel.go
```diff
@@ -79,6 +98,18 @@ func Init(ctx context.Context) func() error {
 		hostName = setHostname()
 	}
 
+	defaultOptions := &initDefaultOptions{
+		defaultAttributes: []attribute.KeyValue{
+			semconv.ServiceName(cfg.ServiceName),
```
brandon-welsch, line 103:
```suggestion
			// https://opentelemetry.io/docs/specs/semconv/resource/#service
			semconv.ServiceName(cfg.ServiceName),
```

Can you set back the doc url, it is kind of useful

otel/otel.go
```diff
@@ -79,6 +98,18 @@ func Init(ctx context.Context) func() error {
...
+			semconv.ServiceInstanceID(serviceInstanceID),
+			semconv.HostName(hostName),
```
brandon-welsch, line 105: Can you set back the doc url, it is kind of useful
```suggestion	
            // https://opentelemetry.io/docs/specs/semconv/resource/host/
			semconv.HostName(hostName),
```

errors/cause_test.go
leo-scalingo: Because CI was broken (lint) and didn't want to loose time in another PR

otel/otel.go
leo-scalingo: Sorry, it was a mistake

otel/otel.go
leo-scalingo: Done

PR #1558, opened 2026-04-09
[storage] Bump v1.8.0

curzolapierre:


matthieu526-scalingo (approved):
We definitely haven't been consistent between all PRs 😅 . Sometimes it is a minor, sometimes a new patch. But anyway it's too late now.

PR #1615, opened 2026-06-03
fix(nsq*producer): correctly declare the v2

EtienneM:
In https://github.com/Scalingo/go-utils/pull/1551 and https://github.com/Scalingo/go-utils/pull/1569, we released the v2 for `nsqproducer` and `nsqlbproducer`. But we forgot to update the major version in the go.mod. This PR aims at fixing this.


nsqlbproducer/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## To be Released
 
+* fix(nsq*producer): correctly declare the v2
```
matthieu526-scalingo: question: should it be
```suggestion
* fix(nsqlbproducer): correctly declare the v2
```
? Or you kept it generic on purpose?

matthieu526-scalingo (commented):
Sorry for that, this is my bad. I still wonder, is it absolutely required? I feel like it is not a widely used practice to update the path of the module to reflect its version number.

nsqlbproducer/CHANGELOG.md
EtienneM: oups, nice catch

matthieu526-scalingo (approved):

PR #1619, opened 2026-06-04
refactor(nsqproducer): replace use of `fmt.Errorf` with `errors.Newf`

EtienneM:

### DIFF
--- a/nsqproducer/producer.go
+++ b/nsqproducer/producer.go
@@ -3,7 +3,6 @@ package nsqproducer
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"time"
 
 	"github.com/gofrs/uuid/v5"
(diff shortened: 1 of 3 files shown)


nsqproducer/producer.go
```diff
@@ -51,10 +50,10 @@ type NsqMessageSerialize struct {
 
 var _ Producer = &NsqProducer{} // Ensure that NsqProducer implements the Producer interface
 
-func New(opts ProducerOpts) (*NsqProducer, error) {
+func New(ctx context.Context, opts ProducerOpts) (*NsqProducer, error) {
 	client, err := nsq.NewProducer(opts.Host+":"+opts.Port, opts.NsqConfig)
 	if err != nil {
-		return nil, fmt.Errorf("init-nsq: cannot initialize nsq producer: %v:%v", opts.Host, opts.Port)
+		return nil, errors.Newf(ctx, "initialize NSQ producer: %v:%v", opts.Host, opts.Port)
```
EtienneM, line 56: We prefer using `github.com/Scalingo/go-utils/errors/v3` rather than the standard library so that we keep the context fields.

briceamen (approved):

PR #1674, opened 2026-08-04
fix(otel) Fix broken mocks and add script to auto-fix them

john-scalingo:
Fix #1673


otel/scripts/postprocess-mocks
EtienneM, line 1: I think it can be a pure Ruby script rather than a weird mix of Bash and Ruby

EtienneM (changes requested):

EtienneM (approved):

PR #1679, opened 2026-08-14
fix(cronsetup): skip missed cron occurrences

### LINKED ISSUE #1646: Prevent cron from running all missed occurrences after a pause
## Problem

When the cron scheduler resumes after an interruption, it immediately runs every occurrence missed during that period. For frequently scheduled jobs, this can cause a large burst of executions.

## Expected behavior

Missed occurrences should be skipped and scheduling should resume from the current time. If some callers rely on catch-up behavior, this could be configurable.

briceamen:
After a pause, cronsetup advanced each entry from its stale scheduled time and replayed every missed slot.

Advance the next run from the timer delivery time instead. A due job runs once after resume, then its next run is scheduled after the current time. The etcd mutex still uses the original scheduled slot.

Closes #1646.

Tests:
- `go test -mod=mod ./...`
- `go test -mod=mod -race ./internal/cron -run '^TestRunEntriesSkipsMissedOccurrences$' -count=20`

### DIFF
--- a/cronsetup/internal/cron/cron.go
+++ b/cronsetup/internal/cron/cron.go
@@ -275,16 +275,7 @@ func (c *Cron) run(ctx context.Context) {
 
 		select {
 		case now = <-time.After(effective.Sub(now)):
-			// Run every entry whose next time was this effective time.
-			for _, e := range c.entries {
-				if e.Next != effective {
-					break
-				}
-				e.Prev = e.Next
-				e.Next = e.Schedule.Next(effective)
-
-				go c.runEntry(ctx, effective, e)
-			}
+			c.runEntries(ctx, effective, now)
 			continue
 
 		case newEntry := <-c.add:
@@ -300,6 +291,19 @@ func (c *Cron) run(ctx context.Context) {
 
 		// 'now' should be updated after newEntry and snapshot cases.
 		now = time.Now().Local()
+	}
+}
+
+func (c *Cron) runEntries(ctx context.Context, effective, now time.Time) {
+	// Run each due entry once and skip any occurrences missed while the scheduler was paused.
+	for _, e := range c.entries {
+		if e.Next != effective {
+			break
+		}
+		e.Prev = e.Next
+		e.Next = e.Schedule.Next(now)
+
+		go c.runEntry(ctx, effective, e)
 	}
 }
 
--- a/cronsetup/internal/cron/cron.go
+++ b/cronsetup/internal/cron/cron.go
@@ -26,6 +26,8 @@ type Cron struct {
 	funcCtx           func(context.Context, Job) context.Context
 	running           bool
 	etcdMutexBuilder  EtcdMutexBuilder
+	timeNow           func() time.Time
+	timeAfter         func(time.Duration) <-chan time.Time
 }
 
 // Job contains 3 mandatory options to define a job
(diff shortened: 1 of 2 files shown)


cronsetup/internal/cron/cron.go
```diff
@@ -303,6 +294,19 @@ func (c *Cron) run(ctx context.Context) {
 	}
 }
 
+func (c *Cron) runEntries(ctx context.Context, effective, now time.Time) {
+	// Run each due entry once and skip any occurrences missed while the scheduler was paused.
+	for _, e := range c.entries {
+		if e.Next != effective {
```
sc-david-voisin, line 306: question: as `e.Next` does not seem set anymore with `effective`, is this test still useful ?

briceamen: Could you clarify what you mean here? The test simulates the scheduler waking up late and checks that missed runs are skipped. Do you think this no longer matches the real flow?

sc-david-voisin: By test, I intended this check:
```
if e.Next != effective {
	break
}
```
which seems related to the former assignation:
```
e.Next = e.Schedule.Next(effective)
```

Now that `e.Next` is set from `e.Schedule.Next(now)`, is this check till useful ? Asking myself if it has not become dead code.

briceamen: Yes, this check is still useful because `effective` is the time when the current job was supposed to run, while `now` is the actual time when the scheduler wakes up.

For example:

- At 09:50, the next jobs are scheduled for 10:00 and 11:00, so `effective` is 10:00
- The scheduler is paused and wakes up at 10:30, so `now` is 10:30 but `effective` is still 10:00
- The loop goes through every job, ordered by `e.Next`
- For the 10:00 job, `e.Next != effective` is false because both values are 10:00, so the job runs
- For the 11:00 job, the condition is true because 11:00 is different from `effective` (10:00), so the loop stops.
- Without this check, nothing would stop the loop before the 11:00 job, and it would run too at 10:30
- The 10:00 job is then rescheduled from `now`, so its next run is 10:31 instead of replaying every missed minute.

I hope I explained this clearly enough 😄️

sc-david-voisin: Clear enough :)
Thanks for the explanation.

sc-david-voisin (approved):

cronsetup/internal/cron/cron.go
```diff
@@ -275,16 +275,7 @@ func (c *Cron) run(ctx context.Context) {
...
-			for _, e := range c.entries {
-				if e.Next != effective {
-					break
-				}
-				e.Prev = e.Next
-				e.Next = e.Schedule.Next(effective)
-
-				go c.runEntry(ctx, effective, e)
-			}
+			c.runEntries(ctx, effective, now)
```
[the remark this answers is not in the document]
briceamen: This is actually a very good finding, the code should refresh the clock explicitly

briceamen:
@sc-david-voisin  I added `timeNow` and `timeAfter` to `Cron` struct so the tests can simulate a delayed timer and cover Copilot finding. The scheduler now
refreshes the current time after the timer fires.

cronsetup/internal/cron/cron.go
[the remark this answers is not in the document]
sc-david-voisin: I agree

sc-david-voisin (approved):

